PR #11, opened 2024-03-19
Dev uniwig

donaldcampbelljr:
Attempt adding uniwig to genimtools.


donaldcampbelljr:
This now supports creating `.npy` files for each chrom and associated meta files for starts, ends, and core:
image

Each meta file holds the header information. Currently, it is the exact same as that which would be placed into the `.wig` file:
image


Above examples created using this test bedfile:
```
chr11 	10	50
chr11	20	76
chr12	769	2395
chr13	771	3000
chr14	800	2900
chr21	1	30
chr21	2	19
chr21	16	31
```

nleroy917:
Now would be a great time to try all this out, because @ClaudeHu needs to create a universe with `uniwig`, but perhaps he should try this implementation first. @donaldcampbelljr Is this ready to be tested?

gtars/src/uniwig/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Note that we provide a chrom.sizes reference file (hg38) in the testing folder -> `genimtools/tests/hg38.chrom.sizes`
+
+### Usage
+```
+Usage: genimtools uniwig --bed <bed> --chromref <chromref> --smoothsize <smoothsize> --stepsize <stepsize> --fileheader <fileheader> --outputtype <outputtype>
+
+Options:
+  -b, --bed <bed>                Path to the combined bed file we want to tranforms
```
nleroy917, line 43: What do you think about making this a positional argument instead?

gtars/tests/test.rs
```diff
@@ -0,0 +1,123 @@
...
+        println!("{:?}", first_line);
+
+        let result = parse_bed_file(&first_line);
+
+        if let Some((ctg, st, en)) = result {
+
+            println!("ctg: {}", ctg);
+            println!("st: {}", st);
+            println!("en: {}", en);
+            assert_eq!(st, 7915738);
```
nleroy917, line 52: Super small thing, but I love to separate large numbers with underscores. So, `7915738`, becomes `7_915_738`. really improves readability.

gtars/tests/test.rs
```diff
@@ -0,0 +1,123 @@
...
+        } else {
+            println!("Failed to parse BED record");
```
nleroy917, line 54: Should the test bail or `panic!` instead?

gtars/tests/test.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+    }
+
+    #[rstest]
+    fn test_read_bed_vec(path_to_bed_file: &str, path_to_bed_file_gzipped: &str) {
+
+        read_bed_vec(path_to_bed_file);
+        read_bed_vec(path_to_bed_file_gzipped);
```
nleroy917, line 63: same here, should we be asserting something?

gtars/src/uniwig/README.md
nleroy917, line 1: Could we put some of this documentation into the actual source code? Using those `///` comment delimiters?

nleroy917: That way it gets pulled into `docs.rs`... I think

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+pub mod cli;
+
+pub mod consts {
+    pub const UNIWIG_CMD: &str = "uniwig";
+
 }
+
+pub struct Chromosome {
+    chrom: String,
+    starts: Vec<i32>,
```
nleroy917, line 21: Should this be `u32` since regions can't have negative start and ends?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+            ends: self.ends.clone(),   // Clone the vector
+        }
+    }
+}
+
+
+// pub fn read_bed_map(combinedbedpath: &str){
+//
+//
+// }
```
nleroy917, line 38: Delete?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+
+pub fn read_bed_vec(combinedbedpath: &str) -> Vec<Chromosome> {
```
nleroy917, line 40: What do you think about having this function return the `Result` enum? So instead, it will return:
```rs
Result<Vec<Chromosome>>
```
Since there are lot of issues that can arise with reading from files....

nleroy917: Also, snake case for this variable? `combinedbedpath` --> `combined_bed_path`

nleroy917: Also also, how about this:
```rs
pub fn read_bed_vec<P: AsRef<Path>>(combinedbedpath: P) -> Vec<Chromosome> { }
```
The trait bounds make it more flexible for what it can accept as input: https://stackoverflow.com/questions/62603012/rewriting-a-function-to-accept-asrefpath-instead-of-path

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+        starts: vec![],
+        ends: vec![],
+    };
+
+    let mut chromosome_vec: Vec<Chromosome> = Vec::new();
+
+    let mut chrom = String::new();
+
+    for line in reader.lines() {
+        //println!("Here is line{:?}", line);
```
nleroy917, line 67: delete?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+
+        // Must use a 2nd let statement to appease the borrow-checker
+        let line_string = line.unwrap();
```
nleroy917, line 70: If we opt to return a `Result` enum, you ca just do:
```rs
let line = line?;
```

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+        }
+
+        chromosome.starts.push(parsed_start);
+        chromosome.ends.push(parsed_end);
+
+    }
+
+    // Is this final sort and push actually necessary?
+    chromosome.starts.sort_unstable();
+    chromosome.ends.sort_unstable();
```
nleroy917, line 108: Also, why unstable? 😵‍💫

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+
+    println!("Reading Bed file complete.");
+
+    //chromosome_vec.sort_by_key(|c| c.chrom.clone());
+
+    return chromosome_vec
+
+}
+
+pub fn parse_bed_file(line: &str) -> Option<(String, i32, i32)> {
```
nleroy917, line 119: Same thing here about the `AsRef<Path>` trait bounds instead of just `&str`

nleroy917: I really like that this returns an `Option`

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+        .get_one::<String>("outputtype")
+        .expect("output type is required");
+
+
+    uniwig_main(*smoothsize, combinedbedpath, chromsizerefpath, bwfileheader, output_type)
+
+
+}
+
+fn clamped_start_position(start:i32, smoothsize: i32) -> i32{
```
nleroy917, line 166: `smooth_size` over `smoothsize`?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+    let mut meta_data_file_names: [String; 3] = ["placeholder1".to_owned(), "placeholder2".to_owned(), "placeholder3".to_owned()];
+
+    // TODO determine potential file types
+    file_names[0] = format!("{}_{}.{}", bwfileheader, "start", output_type);
+    file_names[1] = format!("{}_{}.{}", bwfileheader, "end", output_type);
+    file_names[2] = format!("{}_{}.{}", bwfileheader, "core", output_type);
+
+    meta_data_file_names[0] = format!("{}{}.{}", bwfileheader, "start","meta");
+    meta_data_file_names[1] = format!("{}{}.{}", bwfileheader, "end","meta");
+    meta_data_file_names[2] = format!("{}{}.{}", bwfileheader, "core","meta");
```
nleroy917, line 191: I might be misinterpreting, but is there a reason to store these in an array instead of just creating three new variables for them?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+
+
+    let chrom_sizes = match read_chromosome_sizes(combinedbedpath) {
+        // original program gets chromosome size from a .sizes file, e.g. chr1 248956422
+        // the original program simply pushes 0's until the end of the chromosome length and writes these to file.
+        // can we instead just use the last endsite for each chromosome to save space in th wiggle file?
+
+        Ok(chrom_sizes) => chrom_sizes,
+        Err(err) => {
+            println!("Error reading chromosome sizes: {}", err);
```
nleroy917, line 202: Should we bubble up the error + message?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+
+
+
+
+
+    let chromosomes: Vec<Chromosome> = read_bed_vec(combinedbedpath);
+
+    let num_chromosomes = chromosomes.len();
+
+    println!(" DEBUG Number of Chromosomes{:?}", num_chromosomes);
```
nleroy917, line 215: remove?

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+            file.write_all(b"\n").unwrap();
+            position+=1;
+        }
+
+    }
+
+
+}
+
+fn read_chromosome_sizes(chrom_size_path: &str) -> Result<std::collections::HashMap<String, u32>, Box<dyn Error>> {
```
nleroy917, line 413: I think I have `anyhow` as a dependency in this crate, so you can actually use `anyhow::Result` here instead which is effectively the same thing as `Result<T, Box<dyn Errror>>`

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+    current_end_site = adjusted_start_site + 1 + smoothsize*2;
+
+    //println!("DEBUG: INITIAL ENDSITE -> {}", current_end_site.clone());
+
+    if adjusted_start_site < 1{
+        adjusted_start_site = 1;
+    }
+
+    //println!("DEBUG: SKIPPING UNTIL COORDINATE_POSITION < ADJUSTEDSTARTSITE -> {}  {}", coordinate_position.clone(), adjusted_start_site.clone());
+    while coordinate_position < adjusted_start_site{
```
nleroy917, line 479: Any condition to stop if something goes horribly wrong? haha

gtars/src/uniwig/mod.rs
```diff
@@ -1,16 +1,722 @@
...
+    return (v_coord_counts, v_coordinate_positions)
+}
+#[allow(unused_variables)]
+pub fn fixed_core_wiggle(starts_vector: &Vec<i32>, ends_vector: &Vec<i32>, chrom_size: i32, stepsize:i32) -> (Vec<u32>, Vec<i32>) {
+    // This function is a more direct port of fixedCoreBW from uniwig written in CPP
+    // It allows the user to accumulate reads of across paired starts and ends.
+    // Counts occur between a start coordinate (cutSite) and an end site (endSite) where the endsite is determined based on
+    // the paired ends.
+    // Counts are reported over a stepsize (with a default of stepsize = 1)
+    // Unlike the original function, it does not write to disk in chunks. it simply returns a vector of accumulated reads.
```
nleroy917, line 589: If you move these above the function header... I believe they get pulled into `docs.rs` and will appear in type hints in peoples IDEs

nleroy917 (changes requested):
Looks great! Super great job this will be so useful. I'm sure it works fine, but there are just some cleanup things I think that could be done:

1. Can you run `cargo fmt` in the `gtars` folder?
2. A lot of functions could benefit from an `anyhow::Result<T>` enum I think
3. Many `println!` calls that might look odd when running the `cli`
4. Small naming things for variables (mostly suggesting `snake_case`)
5. Do we plan on making python bindings?

This is so great though -- it was very needed and this will make the code immensely more maintainable.

gtars/src/uniwig/mod.rs
donaldcampbelljr: From docs:

When applicable, unstable sorting is preferred because it is generally faster than stable sorting and it doesn’t allocate auxiliary memory. See [sort_unstable](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable).

https://doc.rust-lang.org/std/primitive.slice.html#method.sort

gtars/src/uniwig/mod.rs
donaldcampbelljr: Some of these items are almost 1 to 1 ports from the c code to help with ensuring I can pick up easily if I step away from it for a while. In this case, we could just create three new variables.

gtars/src/uniwig/README.md
donaldcampbelljr: I like explicit arguments.

gtars/tests/test.rs
donaldcampbelljr: Yeah. This is a good point; I like this approach.

gtars/src/uniwig/mod.rs
donaldcampbelljr: The starts/ends default to -1 if they cannot be parsed:
`let st = fields.next().and_then(|s| s.parse::<i32>().ok()).unwrap_or(-1);`

gtars/src/uniwig/README.md
nleroy917: Ok! Sounds good

gtars/src/uniwig/mod.rs
nleroy917: I see... thats a bummer since its a lot memory allocated for a simple -1 :/

gtars/src/uniwig/mod.rs
nleroy917: I gotcha -- I think thats a valid approach for now haha

gtars/src/uniwig/mod.rs
donaldcampbelljr: On 64 bit architecture, isn't the usize larger than i32?

donaldcampbelljr: Oh never mind, I misread the first comment.

nleroy917:
@donaldcampbelljr @ClaudeHu and I found a small "user experience" thing. Claude was trying to run this with an invalid `combinedbed` file and the error message that we were getting was: `Error reading chromosome sizes: ...`

However, it sent us on the wrong path because the program doesn't seem to use that anyways and just goes for the `combinedbed` argument.

We got it all sorted, but had to follow the breadcrumbs for a bit, so maybe we could update that error message if we choose to stick with keeping that method of identifying chrom sizes.

Also this seems useful: https://stackoverflow.com/questions/42275777/how-to-trace-the-cause-of-an-error-result

nleroy917:
@donaldcampbelljr one more thing...

@ClaudeHu found that when he provided an incorrect `outputtype` argument, it said it was defaulting to `wiggle` but instead did nothing. Looking at this match:
```rust
match output_type { } 
```
It seems like it just does nothing if a non-standard output type is provided, it just prints a misleading statement:
```
_ => {println!("Default to wig file.")},
```
Should it `panic!`?

nleroy917:
One more.....

In `fn write_to_wig_file`: should we create any necessary parent directories for the output file if they don't exist yet?
```rust

    let mut file = OpenOptions::new()
        .create(true)  // Create the file if it doesn't exist
        .append(true)  // Append data to the existing file if it does exist
        .open(filename).unwrap();
```

gtars/tests/test.rs
donaldcampbelljr: Yeah, it should panic. Fixed.

gtars/tests/test.rs
donaldcampbelljr: Yes. Fixed.

gtars/src/uniwig/mod.rs
nleroy917: Maybe we could make use of a library like this: https://github.com/tokio-rs/tracing

donaldcampbelljr:
> @donaldcampbelljr one more thing...
> 
> @ClaudeHu found that when he provided an incorrect `outputtype` argument, it said it was defaulting to `wiggle` but instead did nothing. Looking at this match:
> 
> ```rust
> match output_type { } 
> ```
> 
> It seems like it just does nothing if a non-standard output type is provided, it just prints a misleading statement:
> 
> ```
> _ => {println!("Default to wig file.")},
> ```
> 
> Should it `panic!`?

I've now added it such that it will default to writing `npy` files appropriately.

gtars/src/uniwig/mod.rs
donaldcampbelljr: great idea. Done!

donaldcampbelljr:
> One more.....
> 
> In `fn write_to_wig_file`: should we create any necessary parent directories for the output file if they don't exist yet?
> 
> ```rust
>     let mut file = OpenOptions::new()
>         .create(true)  // Create the file if it doesn't exist
>         .append(true)  // Append data to the existing file if it does exist
>         .open(filename).unwrap();
> ```

Yeah, I solved this in the IGD branch. I've added the functionality here as well. However, I'd like to move this into common functions once both branches have been merged down to dev.

gtars/src/uniwig/mod.rs
donaldcampbelljr: Glancing at this, we could probably break this if we gave it a _negative_ step size, so we could ensure that is not possible.

gtars/src/uniwig/README.md
donaldcampbelljr: added some documentation for example execution

donaldcampbelljr:
Added better error propagation.

I also made the chromsize reference _optional_.

I found that the code was already ignoring the chrom.sizes file and simply using the combined bed file to determine the chromosome size from the last end position of the chromosome.

Now, the user can use either approach. Using the chromsize ref path for smaller bed files with regions far upstream to the chromosome end will be slower as uniwig will pad out the remaining values with 0s until it reaches the end of the chrom.

Currently, it is unclear if we actually _need_ to know the end of the chromosome for uniwig's downstream uses.

PR #36, opened 2024-10-18
Dev narrow peak scoring

donaldcampbelljr:
Add counting based on narrowpeak score.

Can now give uniwig a narrowPeak file and and set score flag via 

```
-o true
```

Also refactored for convenience.


nleroy917:
Just ran it now and it seems to work! Very briefly... I was wondering if we could make that `-o` flag a boolean just by itself. I wonder if saying `-o true` is a tad redundant because we can just say `-o` and that represents true. The absence of the flag, therefore, represents `false`.

nleroy917:
Pending creation of `.bw` files and inspecting in IGV

gtars/src/uniwig/cli.rs
```diff
@@ -70,4 +70,12 @@ pub fn create_uniwig_cli() -> Command {
...
                 .required(false),
         )
+        .arg(
+            Arg::new("score")
+                .long("score")
+                .short('o')
+                .value_parser(clap::value_parser!(bool))
+                .help("Count via score (narrowPeak only!)")
+                .required(false),
+        )
```
nleroy917, line 80: I mentioned already but can we simply make the presence of this flag represent `true`? So that one doesn't need to say `-o true`?

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+        adjusted_start_site.0 = 1;
+    }
+
+    while coordinate_position < adjusted_start_site.0 {
+        // Just skip until we reach the initial adjusted start position
+        // Note that this function will not return 0s at locations before the initial start site
+        coordinate_position = coordinate_position + stepsize;
+    }
+
+    for (index, coord) in starts_vector.iter().enumerate().skip(0) {
```
nleroy917, line 48: Doesnt the `.skip(0)` here just say, "Skip nothing" which is the same as not including it at all?

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
+/// This function is a more direct port of smoothFixedStartEndBW from uniwig written in CPP.
+/// It allows the user to accumulate reads of either starts or ends.
+/// Counts occur between a start coordinate (cutSite) and an end site (endSite) where the endsite is determined based on
+/// the level of smoothing.
+/// counts are reported over a stepsize (with a default of stepsize = 1).
+/// Unlike the original function, it does not write to disk in chunks. it simply returns a vector of accumulated reads.
+#[allow(unused_variables)]
+pub fn start_end_counts(
+    starts_vector: &Vec<(i32, i32)>,
```
nleroy917, line 9: This is requiring an object reference, when a slice type will do:

https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

nleroy917: From that link:

> #### Why this is bad
> Requiring the argument to be of the specific type makes the function less useful for no benefit; slices in the form of &[T] or &str usually suffice and can be obtained from other types, too.

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    let mut coordinate_value: (i32, i32);
+    let mut prev_coordinate_value = 0;
+
+    let mut adjusted_start_site: (i32, i32);
+    let mut current_end_site: (i32, i32);
+
+    let mut collected_end_sites: Vec<(i32, i32)> = Vec::new();
+
+    adjusted_start_site = starts_vector[0].clone(); // get first coordinate position
```
nleroy917, line 31: I think the `.clone()` here is unnecessary because both values in the tuple, `i32`s, implement `Copy`

donaldcampbelljr:
> Just ran it now and it seems to work! Very briefly... I was wondering if we could make that `-o` flag a boolean just by itself. I wonder if saying `-o true` is a tad redundant because we can just say `-o` and that represents true. The absence of the flag, therefore, represents `false`.

Yes. I would prefer that, but CLAP was giving me issues with it last week. I suppose I was just doing it wrong? Can you shed some light on how the argument should be wired up? 
`.default_value(true) or .default_value('true') or .default_value(1)` all seemed to give me a compiler error.

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        while coordinate_position < adjusted_start_site.0 {
+            while current_end_site.0 == coordinate_position {
+                count = count - current_score;
+
+                if count < 0 {
+                    count = 0;
+                }
+
+                if collected_end_sites.last() == None {
```
nleroy917, line 80: I think that a slightly more "rusty" (or idiomatic) way of doing this is to say `collected_end_sites.last().is_none()`

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+        // Apply a bound to push the final coordinates otherwise it will become truncated.
+
+        while current_end_site.0 == coordinate_position {
+            let current_score = adjusted_start_site.1;
+            count = count - current_score;
+            if count < 0 {
+                count = 0;
+            }
+
+            if collected_end_sites.last() == None {
```
nleroy917, line 112: Same here with the `is_none()`

nleroy917:
https://stackoverflow.com/a/60458834/13175187

I think you can set, on your `Arg` struct:
```rust
.num_args(0)
```

nleroy917:
```rust
.arg(
    Arg::with_name("metal")
        .long("metal-micky")
        .required(false)
        .num_arfgs(false)
        .help("I want metal micky"),
)
```

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+/// This function is a more direct port of fixedCoreBW from uniwig written in CPP
+/// It allows the user to accumulate reads across paired starts and ends.
+/// Counts occur between a start coordinate (cutSite) and an end site (endSite) where the endsite is determined based on
+/// the paired ends.
+/// Counts are reported over a stepsize (with a default of stepsize = 1)
+/// Unlike the original function, it does not write to disk in chunks. it simply returns a vector of accumulated reads.
+#[allow(unused_variables)]
+pub fn core_counts(
+    starts_vector: &Vec<(i32, i32)>,
+    ends_vector: &Vec<(i32, i32)>,
```
nleroy917, line 140: Same here with the slices over object references. I.e. `&[(i32, i32)]`

gtars/src/uniwig/counting.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    let mut coordinate_value: (i32, i32);
+    let mut prev_coordinate_value = 0;
+
+    let mut current_start_site: (i32, i32);
+    let mut current_end_site: (i32, i32);
+
+    let mut collected_end_sites: Vec<(i32, i32)> = Vec::new();
+
+    current_start_site = starts_vector[0].clone(); // get first coordinate position
```
nleroy917, line 159: I also think this clone might not be needed

gtars/src/uniwig/reading.rs
```diff
@@ -0,0 +1,295 @@
...
+        if chrom.is_empty() {
+            // Initial chromosome
+            chromosome.chrom = String::from(parsed_chr.trim());
+            chrom = String::from(parsed_chr.trim());
+            chromosome.starts.push((parsed_start, default_score));
+            chromosome.ends.push((parsed_end, default_score));
+            continue;
+        }
+
+        if String::from(parsed_chr.trim()) != chrom {
```
nleroy917, line 57: This is creating an owned instance just for comparison. You can use dereferencing:
```rust
if *parsed_chr.trim() != chrom {
```

gtars/src/uniwig/reading.rs
```diff
@@ -0,0 +1,295 @@
...
+        if chrom.is_empty() {
+            // Initial chromosome
+            npchromosome.chrom = String::from(parsed_chr.trim());
+            chrom = String::from(parsed_chr.trim());
+            npchromosome.starts.push((parsed_start, parsed_score));
+            npchromosome.ends.push((parsed_end, parsed_score));
+            continue;
+        }
+
+        if String::from(parsed_chr.trim()) != chrom {
```
nleroy917, line 132: Same here with the owned instance just for comparison

gtars/src/uniwig/reading.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    let mut chromosome_vec: Vec<Chromosome> = Vec::new();
+
+    let mut chrom = String::new();
+
+    for line in reader.lines() {
+        //println!("Here is line{:?}", line);
+
+        // Must use a 2nd let statement to appease the borrow-checker
+        let line_string = line.unwrap();
```
nleroy917, line 43: Is there a way to more gracefully handle this?

gtars/src/uniwig/writing.rs
```diff
@@ -0,0 +1,117 @@
+use crate::uniwig::Chromosome;
+use ndarray::Array;
+use ndarray_npy::write_npy;
+use std::fs::{create_dir_all, remove_file, File, OpenOptions};
+use std::io;
+use std::io::{BufWriter, Write};
+
+pub fn write_to_npy_file(
+    counts: &Vec<u32>,
```
nleroy917, line 9: Slice

gtars/src/uniwig/writing.rs
```diff
@@ -0,0 +1,117 @@
...
+        .unwrap();
+
+    // The original wiggle file header. This can be anything we wish it to be. Currently space delimited.
+    let mut wig_header = "fixedStep chrom=".to_string()
+        + chromname.as_str()
+        + " start="
+        + start_position.to_string().as_str()
+        + " step="
+        + stepsize.to_string().as_str();
+    wig_header.push_str("\n");
```
nleroy917, line 42: Since `\n` is a `char`. It might be better to change to `push('\n')` (note the single quotes

gtars/src/uniwig/writing.rs
```diff
@@ -0,0 +1,117 @@
...
+    file.write_all(wig_header.as_ref()).unwrap();
+}
+
+pub fn write_combined_wig_files(
+    location: &str,
+    output_type: &str,
+    bwfileheader: &str,
+    chromosomes: &Vec<Chromosome>,
```
nleroy917, line 50: Slice

gtars/src/uniwig/mod.rs
```diff
@@ -219,7 +141,9 @@ pub fn uniwig_main(
     // Determine File Type
     let ft = FileType::from_str(filetype.to_lowercase().as_str());
 
-    let stepsize = 1;
+    let score = score;
```
nleroy917, line 144: Isnt this a redundant?

gtars/src/uniwig/mod.rs
```diff
@@ -219,7 +141,9 @@ pub fn uniwig_main(
...
+
+    let stepsize = stepsize;
```
nleroy917, line 146: Oh this too

gtars/src/uniwig/mod.rs
```diff
@@ -183,6 +95,12 @@ pub fn run_uniwig(matches: &ArgMatches) {
         .get_one::<i32>("threads")
         .expect("requires integer value");
 
+    let score = matches.get_one::<bool>("score").unwrap_or_else(|| &false);
```
nleroy917, line 98: Why does `&false` need to be a reference? Arn't booleans a type that implements `Copy`?

nleroy917 (approved):
Some comments here and there but it looks good! I also really like the refactor

donaldcampbelljr:
This one worked:
https://stackoverflow.com/a/77643381

```
.arg(
    Arg::new("metal")
        .long("metal-micky")
        .short('m')
        .action(ArgAction::SetTrue)
        .help("I want metal micky"),
)
```

gtars/src/uniwig/cli.rs
donaldcampbelljr: Done.

gtars/src/uniwig/mod.rs
donaldcampbelljr: Removed them.

gtars/src/uniwig/mod.rs
donaldcampbelljr: Compiler complains. Documentation shows using integers without reference but I dont see a reason for booleans.
image

gtars/src/uniwig/counting.rs
donaldcampbelljr: Done

gtars/src/uniwig/writing.rs
donaldcampbelljr: Yeah. That worked!

gtars/src/uniwig/counting.rs
donaldcampbelljr: Done.

PR #44, opened 2024-10-31
attempt txt input update

sanghoonio:
This is to allow create_igd_f() to work with a .txt input file instead of a directory for fileList of beds.


nleroy917:
Oh so this will basically say:

If `filelist` is an actual file on disk, read it in... otherwise assume its a directory and loop through

nleroy917:
I feel like it should be better to just have it read from `stdin`, right? That way you can be like:
```
cat filelist.txt | gtars igd
```
or
```
ls path/to/files/*.bed.gz | gtars igd
```

sanghoonio:
For usability yeah, just added the txt file input for now because it seemed easier to add. I'll try stdin inputs

donaldcampbelljr:
> I feel like it should be better to just have it read from `stdin`, right? That way you can be like:
> 
> ```
> cat filelist.txt | gtars igd
> ```
> 
> or
> 
> ```
> ls path/to/files/*.bed.gz | gtars igd
> ```

The C version builds from a directory or a .txt file which is why I suggested it (to continue towards total feature parity). Is reading from stdin objectively better?

sanghoonio:
From an R bindings perspective having to write a txt file from R to read in rust might negligibly slow things down over passing in bed filepaths directly. With stdin though, we can input multiple paths like below right? 

```
path/to/files/file1.bed.gz path/to/files/file2.bed.gz path/to/files/file3.bed.gz | gtars igd
```

donaldcampbelljr:
> From an R bindings perspective having to write a txt file from R to read in rust might negligibly slow things down over passing in bed filepaths directly. With stdin though, we can input multiple paths like below right?
> 
> ```
> path/to/files/file1.bed.gz path/to/files/file2.bed.gz path/to/files/file3.bed.gz | gtars igd
> ```

Ah ok. Got it. Thanks.

nleroy917:
> The C version builds from a directory or a .txt file which is why I suggested it (to continue towards total feature parity). Is reading from stdin objectively better?

I'm not too sure which is better. I think its more in line with "UNIX philosophy" to be able to pipe things around, but I suppose I'm not too opinionated one way or another.

donaldcampbelljr:
I fixed this and the functionality now works for reading a .txt file.

donaldcampbelljr:
And I added stdin as an option so you can do this:

```
echo "/home/drc/GITHUB/gtars/gtars/tests/data/igd_file_list/igd_bed_file_1.bed" | cargo run igd create --output /home/drc/IGD_TEST/output/ --filelist -

```

And it works. 

Please confirm this meets your needs and then merge to dev when you are ready.

donaldcampbelljr (approved):
Just make sure this works like you expect it to.

sanghoonio:
> And I added stdin as an option so you can do this:
> 
> ```
> echo "/home/drc/GITHUB/gtars/gtars/tests/data/igd_file_list/igd_bed_file_1.bed" | cargo run igd create --output /home/drc/IGD_TEST/output/ --filelist -
> ```
> 
> And it works.
> 
> Please confirm this meets your needs and then merge to dev when you are ready.


Looks like it creates a .igd with .txt input but it the .igd output doesn't seem to work with search. A .igd created with a bed file directory doesn't work with search either unless I switch back to dev and recompile. Can you let me know if you can reproduce this?

```
sam@Sams-MacBook-Pro release % ./gtars igd create --output /Users/sam/Documents/Work/episcope/.test/igd/ --filelist /Users/sam/Documents/Work/episcope/.test/bed_paths_10.txt

Number of Bed Files found:
10
Temporary Tiles:
 nCtgs (igd.nctg): 63, nRegions (igd.total): 4840383, nTiles (nt): 191146
IGD saved to: /Users/sam/Documents/Work/episcope/.test/igd/igd_database.igd
Total Intervals: 4644521, l_avg: 690.1135
nctg:63  nbp:16384
sam@Sams-MacBook-Pro release % ./gtars igd search --database /Users/sam/Documents/Work/episcope/.test/igd/igd_database.igd --query /Users/sam/Documents/Work/episcope/.test/bed2.bed

thread 'main' panicked at src/igd/search.rs:434:41:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

donaldcampbelljr:
I can't reproduce the above error. However, when using the dev branch of IGD, I notice a bug where I'm creating an .igd from a directory and my search is returning 0 hits.

However, if I create an igd from a single source file and query that source file against the igd, it appears to be working fine.

I will have to investigate this early next week. I suspect the issues may be related.

sanghoonio:
Hmm I'm still getting the same error after pulling and recompiling, are there differences between mac and linux when compiling rust that can cause something like this?

donaldcampbelljr:
Since this functionality works and it is actually igd that needs to be tested further, I'm going to go ahead and merge this work to dev so it does not become stale.

PR #47, opened 2024-11-15
Refactor bam to bw workflow

donaldcampbelljr:
Allows for streaming BAM seq reads to bigwig as well as Bed outputs. 

This will allow us to use gtars in the PEPATAC pipeline.


nleroy917 (approved):
I know this needs to be a quick turnaround, so I just made small comments here and there and didn't go through with particular detail

PR #50, opened 2024-12-05
Add bamshift flag

donaldcampbelljr:
I realized that position shifting is necessary for all bam workflows not just bam ->bed. Therefore, this PR adds a flag that can switch to and from using a position shifting function for the bam workflow. Instead of creating start,end, and core files, it will create a singular shift file.

### DIFF
--- a/gtars/src/uniwig/counting.rs
+++ b/gtars/src/uniwig/counting.rs
@@ -1316,7 +1316,7 @@ pub fn variable_shifted_bam_to_bw( records: &mut Box<Query<noodles::bgzf::reader
 
         let shifted_pos = get_shifted_pos(flags, start_site, end_site);
 
-        let mut adjusted_start_site = shifted_pos - smoothsize;
+        adjusted_start_site = shifted_pos - smoothsize;
 
 
         count += 1;
--- a/gtars/src/uniwig/counting.rs
+++ b/gtars/src/uniwig/counting.rs
@@ -1231,6 +1231,187 @@ pub fn bam_to_bed_no_counts(
     Ok(())
 }
 
+pub fn variable_shifted_bam_to_bw( records: &mut Box<Query<noodles::bgzf::reader::Reader<std::fs::File>>>,
+                               chrom_size: i32,
+                               smoothsize: i32,
+                               stepsize: i32,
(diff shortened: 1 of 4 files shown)


gtars/src/uniwig/counting.rs
```diff
@@ -1319,3 +1464,48 @@ fn set_up_file_output(
         // write to std_out, this will be useful for sending input to bigtools to create bw files
     }
 }
+
+pub fn get_shifted_pos(flags: Flags, start_site:i32, end_site:i32) -> i32 {
```
nleroy917, line 1468: Do we really need to pass `flags` like this? This will take ownership of `flags` and I wonder if borrowing (pass by reference) would be better? (i.e. `&Flags`)

nleroy917 (changes requested):
Just two small things to clarify

gtars/src/uniwig/README.md
```diff
@@ -54,11 +54,11 @@ Options:
   -u, --counttype <counttype>    Select to only output start, end, or core. Defaults to all. [default: all]
   -p, --threads <threads>        Number of rayon threads to use for parallel processing [default: 6]
   -o, --score                    Count via score (narrowPeak only!)
+  -a, --bamshift                 Set bam shift to False, i.e. uniwig will count raw reads without considering read direction.
```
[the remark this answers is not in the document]
donaldcampbelljr: Yeah, I originally had this as --no-bamshift and went back and forth on it. I decided that that the flag is clarified by the text in CLAP. Do you think this is a big deal? I figured most use-cases will _not_ use this flag.

gtars/src/uniwig/counting.rs
donaldcampbelljr: I'll try this.

gtars/src/uniwig/README.md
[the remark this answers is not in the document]
donaldcampbelljr: Ok this is done

gtars/src/uniwig/counting.rs
donaldcampbelljr: This worked and is done.

PR #58, opened 2024-12-17
add ga4gh refget digest functionality

nsheff:
Adds a new `digests` module that provides md5 and ga4gh digest computation for strings and fasta files.

I did not implement a CLI for it, just rust library functions and python bindings. 

I confirmed the python bindings work for me locally.


nsheff:
Do you know why I can't do this?

```
from gtars.digests import digest_fasta
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'gtars.digests'
```

When I *can* do this?
```
from gtars import digests
>>> digests
<module 'digests'>
>>> digests.digest_fasta()
```

nleroy917:
Two things:
1. When I test, I make sure that I am in a different directory, otherwise python can be weird about imports (because there is a `gtars` folder already there so it tries that first...)
2. You might need to create a `bindings/python/gtars/digests/__init__.py` file with this inside:
```python
from .gtars.digests import *
```
You can place a `__init__.pyi` file next to it to add type-annotations/docs for intellisense inside IDE's too. Check out the other folders for examples. Tokenizers is probably a good example. 

The reason this `__init__.py` file is needed is not totally clear to me... but its what I got to work after looking at other examples of Rust + Python bindings for `pyo3` with a module system and then also looking at the [`pyo3` discussion forum](https://github.com/PyO3/pyo3/discussions/3591)

nsheff:
yeah I had figured that out, but even after adding that I still get the same issue...

nsheff:
But now the error is different:

```
from gtars.digests import md5_digest
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nsheff/.local/lib/python3.12/site-packages/gtars/digests/__init__.py", line 1, in <module>
    from .gtars.digests import *  # noqa: F403
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'gtars.digests.gtars'
```

nleroy917:
I think we are missing `sys_modules.set_item("")`... line 35 in `bindings/python/src/lib.rs`

```rust
    // set names of submodules
    sys_modules.set_item("gtars.tokenizers", m.getattr("tokenizers")?)?;
    sys_modules.set_item("gtars.ailist", m.getattr("ailist")?)?;
    sys_modules.set_item("gtars.utils", m.getattr("utils")?)?;
    sys_modules.set_item("gtars.models", m.getattr("models")?)?;
    // <--------    add your module here? 
```

nsheff:
Yep that was what I was missing. it works now! thanks.

nleroy917:
Do we want to merge this soon?

gtars/src/digests/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    if file_path == "-" {
+        Ok(Box::new(std::io::stdin()) as Box<dyn Read>)
+    } else if file_path.ends_with(".gz") {
+        let file = File::open(file_path)?;
+        Ok(Box::new(MultiGzDecoder::new(file)) as Box<dyn Read>)
+    } else {
+        let file = File::open(file_path)?;
+        Ok(Box::new(file) as Box<dyn Read>)
+    }
+}
```
donaldcampbelljr, line 87: We have something similar in utils, should we try to use that instead? It returns a Reader though not a Read.
https://github.com/databio/gtars/blob/bfcabdc8511b0bd1674759dfc6ace847fb5157db/gtars/src/common/utils.rs#L21-L33

donaldcampbelljr (approved):
Great. Thanks for adding tests and docs, too.

gtars/src/digests/mod.rs
nleroy917: Yeah I was going to mention that we have a generic utility for this I tink

nsheff: hm; I'd have to look into what are the differences between a `Read` and a `Reader` -- it it looks like yours is also returning a `Read` object....
One uses `MultiGzDecoder` and the other is `GzDecoder`

nsheff: Ok I integrated these.

Issue #45, opened 2024-11-04
IGD Search is returning 0 hits in some test cases


donaldcampbelljr:
When using the dev branch of IGD, I notice a bug where I'm creating an .igd from a directory and my search is returning 0 hits.

However, if I create an igd from a single source file and query that source file against the igd, it appears to be working fine.

_Originally posted by @donaldcampbelljr in https://github.com/databio/gtars/issues/44#issuecomment-2454841524_

sanghoonio:
Findings from debugging session today:

1. some bed files result in negative avg size when making a .igd
2. for the issue involving index out of bounds: involves gData length in get_overlaps()

Test bed files for making the igd are here: [https://myuva-my.sharepoint.com/:u:/g/personal/sp5fd_virginia_edu/ESZDBr8_mjBDjl9Zr8xAkt8BJKVjoMecpdrGZE3Dqg6Bog?e=BKxNR8](url)

sanghoonio:
Testing with small bed files shows some consistent differences between the Rust and C versions of IGD search (at least with these test files):

1. Rust version does not capture hits at index 0
2. Rust version does not match the number of hits in C version or GenomicRanges::countOverlaps(), and often just returns 1 hit per index

![image](https://github.com/user-attachments/assets/24aa4645-1feb-4656-a865-d6cccf45f1d6)

PR #60, opened 2024-12-18
Dev igd 45

donaldcampbelljr:
Work towards solving #45 as well as adding better test coverage for IGD workflow


donaldcampbelljr:
This should fix issue #45. I also spent time going start to finish using igd create and then search and re-doing the tests to ensure that we have better coverage on both the create and search workflows.

gtars/src/igd/create.rs
```diff
@@ -100,11 +103,11 @@ pub fn igd_get_create_matches(matches: &ArgMatches) {
         .get_one::<String>("dbname")
         .expect("File list path is required");
 
-    create_igd_f(output_path, filelist, db_output_name);
+    let _igd = create_igd_f(output_path, filelist, db_output_name);
 }
 
 /// Creates IGD database from a directory of bed files.
-pub fn create_igd_f(output_path: &String, filelist: &String, db_output_name: &String) {
+pub fn create_igd_f(output_path: &String, filelist: &String, db_output_name: &String) -> igd_t {
```
nleroy917, line 110: Is `igd_t` a struct? What do you think about capitalizing it to be more "Rusty"?

nleroy917 (approved):
Yeah this looks good, maybe just run `cargo fmt`?

gtars/src/igd/create.rs
donaldcampbelljr: At some point we can do a large refactor to tackle things like this. For example, some variables are named in a way that are not super clear. This is because they were ported directly from C code and I wanted to keep them aligned for troubleshooting purposes.

donaldcampbelljr:
We can run cargo fmt on dev right before today's release.

PR #70, opened 2025-01-21
potential solution for skipping sort

donaldcampbelljr:
@ClaudeHu try this branch and see if it:
a. shortens time significantly,
b. provides the same answer for your presorted bed files

The original C++ implementation of uniwig does sort the starts and ends in memory regardless of the sort flag being set.

### DIFF
--- a/gtars/src/uniwig/cli.rs
+++ b/gtars/src/uniwig/cli.rs
@@ -95,6 +95,13 @@ pub fn create_uniwig_cli() -> Command {
                 .action(ArgAction::SetTrue),
         )
         .arg(
+            Arg::new("skip-sort")
+                .long("skip-sort")
+                .short('k')
+                .help("Skip sorting starts and ends during reading of bed files.")
+                .action(ArgAction::SetTrue),
+        )
+        .arg(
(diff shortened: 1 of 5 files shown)


ClaudeHu:
It greatly improved running time

```
real	13m21.378s
user	14m30.424s
sys	2m28.003s
```

donaldcampbelljr:
Just to clarify after speaking with Claude, two things improved the time:
1. building with the release flag
2. re-sorting the bed file before using
I will close this PR and Branch for now as it is probably unecessary

nsheff:
two questions:

1. why should re-sorting the bed file before using it have changed anything, if the sort is supposedly happening internally as well?
2. Would it be helpful if uniwig prompted the user or gave a warning if the file was out of order? (seems like it should be a moot point, but I'm confused)

nleroy917:
Curious if we could complete the experiment (if its even worth it):
|          | No-Release | Release |
|----------|------------|---------|
| Unsorted |            |         |
| Sorted   |            |         |

Fill in the values here to figure out what was the important factor

Issue #90, opened 2025-02-14
Allow writing chrom names with `.` in the name


donaldcampbelljr:
> Ok, I was able to reproduce (was still using a smaller subset of chroms).
> 
> Added a print function immediately before opening each input file for the merge.
> 
> /home/drc/Downloads/gtars_74_issue/output/**testchrY_NC_051844.1_shift.bed**
> thread 'main' panicked at src/uniwig/writing.rs:84:49:
> called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> 
> Checking the folder, I see a file: `testchrY_NC_051844.bed`. Checking chrom.sizes:
> ```
> chrX	124992030
> chrY_NC_051844.1	3937623
> chrY_unplaced_NW_024010443.1	1569522
> ```
> So, it appears to me that it is dropping the `.1` when writing that particular chromosome to file.
> 
> Similar issue for the following chromosome: `chrY_unplaced_NW_024010443.bed` exists without the `.1`. 

 _Originally posted by @donaldcampbelljr in [#74](https://github.com/databio/gtars/issues/74#issuecomment-2660332306)_

nsheff:
You just need to make sure it follows the SAM spec for what's allowed in a chrom name. periods are allowed.

donaldcampbelljr:
One issue:
SAM allows `/` but this currently causes issues when attempting to create temp files before merging.

nleroy917:
I think slashes are very strictly forbidden in file names/systems

nsheff:
yes, you will probably need to accommodate that, then.

Issue #89, opened 2025-02-14
Allow skipping files that do not exist during merge.


donaldcampbelljr:
> @nleroy917 Good idea. By the time the program gets this far, it has checked that the chromosomes exist in the bam file. However, it could raise an error during writing to the bed file line by line, but we _should_ see this error in the terminal (which I don't in this case). We could skip that file. As long as the program communicates that it was expecting a chromosome file but didn't find one that would be fine. Currently, the messaging is inadequate either way as I'm not sure which chrom failed from the backtrace. :) Also, I would think the consumer handle would produce an empty file either way by this point. 

 _Originally posted by @donaldcampbelljr in [#74](https://github.com/databio/gtars/issues/74#issuecomment-2660283185)_

PR #92, opened 2025-02-18
0.2.2 release

donaldcampbelljr:
## [0.2.2]
- fix [#90](https://github.com/databio/gtars/issues/90)
- fix [#89](https://github.com/databio/gtars/issues/89)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,9 +36,9 @@ jobs:
           sccache: 'true'
           manylinux: auto
       - name: Upload wheels
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: wheels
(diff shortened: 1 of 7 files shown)


donaldcampbelljr:
@nleroy917 ready for merge to master and new point release when it is convenient for you.

nleroy917:
Just need to bump version in bindings I think (again, even though it doesn't matter). I feel like we should keep in sync? Maybe not

donaldcampbelljr:
yeah, absolutely. good catch.

nleroy917:
But, do we want to keep them in sync? Not sure if it matters that much -- I've just thus-far kept them in sync

donaldcampbelljr:
I would so as to cut down on any potential confusion.

nleroy917:
Definitely agree with that

Issue #105, opened 2025-03-05
Get rid of warnings


nleroy917:
My local environment (`rust-analyzer` + `clippy`) shows over 170 "problems"/warnings in the codebase. These can be small formatting/idiomatic style changes proposed by clippy down to things like a variable being `mut` when it doesn't need to be.

It would be nice to get rid of these and suppress `clippy` warnings when it makes sense because it makes it harder to debug and detect issues when there are so many of these small warnings.

PR #109, opened 2025-03-20
Remove clippy warnings

nleroy917:
This is a fix for #105 .

I use [`clippy`](https://github.com/rust-lang/rust-clippy) to improve my rust code, catch small things, etc. The current code base had >180 warnings that made it difficult for me to see true errors through the sea of warnings...

I want in and remove them by either 1) implementing the fix, or 2) manually suppressing the warning. Overwhelmingly, the warnings were small things like:
```diff
- cnt = cnt + 1;
+ cnt += 1;
```
or
```diff
- if obj == None
+ if obj.is_none()
```
There were some more elusive warnings about us passing references that get auto-dereferenced by the compiler.

All tests still pass.

### DIFF
--- a/bindings/python/src/tokenizers/py_tokenizers/mod.rs
+++ b/bindings/python/src/tokenizers/py_tokenizers/mod.rs
@@ -1,4 +1,3 @@
-use gtars::tokenizers::config::TokenizerConfig;
 use pyo3::prelude::*;
 use pyo3::types::PyType;
 
--- a/bindings/python/src/tokenizers/py_tokenizers/mod.rs
+++ b/bindings/python/src/tokenizers/py_tokenizers/mod.rs
@@ -178,7 +178,7 @@ impl PyTokenizer {
 
     pub fn __repr__(&self) -> String {
         format!(
-            "TreeTokenizer({} total regions)",
+            "Tokenizer({} total regions)",
             self.tokenizer.get_vocab_size()
         )
     }
--- a/bindings/python/src/tokenizers/py_tokenizers/mod.rs
+++ b/bindings/python/src/tokenizers/py_tokenizers/mod.rs
@@ -133,20 +133,14 @@ impl PyTokenizer {
     pub fn convert_id_to_region(&self, id: u32) -> Option<PyRegion> {
         Python::with_gil(|py| {
             let region = self.universe.borrow(py).convert_id_to_region(id);
-            match region {
-                Some(r) => Some(r.into()),
-                None => None,
-            }
+            region
         })
     }
 
     pub fn convert_region_to_id(&self, region: &PyRegion) -> Option<u32> {
         Python::with_gil(|py| {
             let id = self.universe.borrow(py).convert_region_to_id(region);
-            match id {
-                Some(i) => Some(i),
-                None => None,
-            }
+            id
         })
     }
 
--- a/bindings/python/src/utils/mod.rs
+++ b/bindings/python/src/utils/mod.rs
@@ -22,7 +22,7 @@ pub fn extract_regions_from_py_any(regions: &Bound<'_, PyAny>) -> Result<RegionS
 
         let regions = gtars::common::models::RegionSet::try_from(regions);
         match regions {
-            Ok(regions) => return Ok(RegionSet::from(regions)),
+            Ok(regions) => return Ok(regions),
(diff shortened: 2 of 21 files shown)


bindings/python/src/tokenizers/py_tokenizers/mod.rs
```diff
@@ -134,20 +133,14 @@ impl PyTokenizer {
     pub fn convert_id_to_region(&self, id: u32) -> Option<PyRegion> {
         Python::with_gil(|py| {
             let region = self.universe.borrow(py).convert_id_to_region(id);
-            match region {
-                Some(r) => Some(r.into()),
-                None => None,
-            }
+            region
```
donaldcampbelljr, line 136: Ok, for items like this, are we **certain** region will be `None` without the match statement?

gtars/src/digests/mod.rs
```diff
@@ -120,9 +120,9 @@ pub fn digest_fasta<T: AsRef<Path>>(file_path: T) -> Result<Vec<DigestResult>> {
         let md5 = format!("{:x}", md5_hasher.finalize_reset());
         results.push(DigestResult {
             id: id.to_string(),
-            length: length,
+            length,
             sha512t24u: sha512,
-            md5: md5,
+            md5,
```
donaldcampbelljr, line 125: I believe I see why clippy made this change (due to the redundancy of `length: length`) but at first glance this looks `off` to change this.

gtars/src/igd/create.rs
```diff
@@ -430,7 +429,7 @@ pub fn igd_save_db(igd: &mut igd_t, output_path: &String, db_output_name: &Strin
     }
 
     for i in 0..igd.nctg {
-        let idx = i.clone() as usize;
+        let idx = i as usize;
```
donaldcampbelljr, line 432: Interesting, I feel I cloned this because it was causing me issues (because I'm using it as an index later). BUT assigning a variable to a usize _should_ clone automatically, so I think this is ok. IGD has test coverage so this should be good to go.

donaldcampbelljr (approved):
I also tested changes locally and it passes tests.

bindings/python/src/tokenizers/py_tokenizers/mod.rs
nleroy917: So I think that the reason Clippy sked for this is that the match is redundant... if it matches `Some,` we immediately turn into `Some` again... If it matches `None` we immediately return `None`. I think that's why it had me change this.

bindings/python/src/utils/mod.rs
```diff
@@ -22,7 +22,7 @@ pub fn extract_regions_from_py_any(regions: &Bound<'_, PyAny>) -> Result<RegionS
 
         let regions = gtars::common::models::RegionSet::try_from(regions);
         match regions {
-            Ok(regions) => return Ok(RegionSet::from(regions)),
+            Ok(regions) => return Ok(regions),
```
khoroshevskyi, line 25: Is it correct change?. It seems it does nothing

gtars/src/digests/mod.rs
nleroy917: Yeah JavaScript does this too. But since we are not doing it for ALL fields, we can suppress the warning as I see the argument that it looks off to mix that behavior

gtars/src/common/models/region_set.rs
```diff
@@ -461,7 +462,7 @@ mod tests {
 
         let tempdir = tempfile::tempdir().unwrap();
 
-        let mut new_file_path = PathBuf::try_from(tempdir.into_path()).unwrap();
+        let mut new_file_path = tempdir.into_path();
```
khoroshevskyi, line 465: Really nice!

gtars/src/digests/mod.rs
nleroy917: Ok I went in and suppressed this warning since I kinda agree it should be one way or the other

khoroshevskyi (approved):
Everything looks good!!

bindings/python/src/utils/mod.rs
nleroy917: True -- I agree I dont know let me look into this

nleroy917: Yeah I can break early using the `?` operator:
```rust
let regions = gtars::common::models::RegionSet::try_from(regions)?;
return Ok(regions);
```

PR #111, opened 2025-03-21
Added calculation of file_digest to RegionSet struct

khoroshevskyi:
We would like to calculate bed file digest of whole file, not just region. In this PR you can find new function that calculates this digest.

p.s.
I am sorting bed file before calculation of the digest, and because of that I am cloning RegionSet object. Is there a better way of doing it?


nleroy917 (approved):
Looks good

donaldcampbelljr (approved):
Looks good!

PR #113, opened 2025-04-04
Fixed to_bigbed for unsorted files. New bed identifier

khoroshevskyi:


gtars/src/common/models/region_set.rs
```diff
@@ -315,7 +317,8 @@ impl RegionSet {
     /// - out_path: the path to the bigbed file which should be created
     /// - chrom_size: the path to chrom sizes file
     ///
-    pub fn to_bigbed<T: AsRef<Path>>(&self, out_path: T, chrom_size: T) -> Result<()> {
+    pub fn to_bigbed<T: AsRef<Path>>(&mut self, out_path: T, chrom_size: T) -> Result<()> {
+        self.sort();
```
nleroy917, line 321: Should we add something to the docs string of this saying that `to_bigbed` might mutate/sortthe underlying regions?

PR #122, opened 2025-05-12
Incorrect header parsing

khoroshevskyi:


gtars/src/common/models/region_set.rs
```diff
@@ -66,10 +68,26 @@ impl TryFrom<&Path> for RegionSet {
                 | string_line.starts_with("#")
             {
                 header.push_str(&string_line);
+                first_line = false;
                 continue;
             }
 
-            // println!("parts: {:?} -- {:?}", parts, string_line);
+            // Handling headers like `chr start end etc` without #
```
donaldcampbelljr, line 75: what about multiple lines? And lines that do begin with `#`?

donaldcampbelljr (commented):
this seems fine but it is only a specific case...shouldn't we generalize?

gtars/src/common/models/region_set.rs
khoroshevskyi: we already have it. It is being checked before this check

donaldcampbelljr: Ok, so this is really a 2nd check, and it is specifically for a _column header_.
```suggestion
            // Handling column headers like `chr start end etc` without #
```

donaldcampbelljr: Curious, do we have an idea of what percentage of files this is occurring for that we have encountered so far?

Issue #72, opened 2025-01-24
Add ability to sort then merge multiple files


donaldcampbelljr:
- [x] need to be able to sort and then merge bed files (see this example code for sorting and merging: https://github.com/databio/uniwig/blob/dev/old/analysis.sh)

donaldcampbelljr:
- [ ] We'd like to have this added before publication.

PR #131, opened 2025-06-30
Use uniwig on a directory of bed files and/or narrowPeak files

donaldcampbelljr:
Allows the user to point uniwig at a directory of bed files and/or narrowPeak files.

Also enforces the use of a chrom.sizes reference file (instead of inferring from the singular bed file).


donaldcampbelljr:
also addressed #72

gtars/tests/test.rs
```diff
@@ -725,6 +719,106 @@ mod tests {
...
+            filetype,
+            num_threads,
+            score,
+            stepsize,
+            zoom,
+            false,
+            true,
+            1.0,
+        )
+            .expect("Uniwig main failed!");
```
ClaudeHu, line 818: I was a little bit confused here, I thought contents are written to `tempchromsizerefpath` during the test, but didn't see code that writes to this path.

ClaudeHu (approved):

PR #138, opened 2025-07-20
New overlappers and tokenizer structure

nleroy917:
### Overview
We discussed designating `gtars` as a central location for a few interval overlap tools, namely [BITS](https://arxiv.org/abs/1208.3407) and [AIList](https://www.biorxiv.org/content/10.1101/593657v1). Currently, we are importing @sstadick's (amazing) [`rust_lapper`](https://github.com/sstadick/rust-lapper), but it was discussed that having the actual code live in `gtars` next to other overlap implementations like AIList would be pretty nice -- especially for the paper we plan to write.

## Implementations
To that end, I created a new module in this PR called `overlap`. This module has a trait called `Overlapper` which requires two methods:
1. `build`: simply be able to build a data structure from a set of intervals, and
2. `find`: perform the interval overlap computation

I then brought over BITS from `rust_lapper` and added our implementation of AIList and then implemented `Overlapper` for both of these structs.

## Tokenizers
With this new structure, I removed some complexity from the tokenizers and switched it so that anything implementing `Overlapper` can now be used as a tokenizer "backend"
```rust
pub struct Tokenizer {
    core: HashMap<String, Box<dyn Overlapper<u32, u32>>>,
    universe: Universe,
    special_tokens: SpecialTokens, 
}
```


nleroy917:
@sstadick can't figure out how to request you as a reviewer on this. As mentioned via email, I would love to find a way to make it so that the `Overlapper` trait requires you to return an iterator (sorta like `IterFind`). But, I couldn't get that working.

PR #139, opened 2025-07-21
Minor change to allow Python bingding RegionSet init from list of Region

ClaudeHu:


donaldcampbelljr (approved):
Looks good. Just had a question about collecting.

PR #141, opened 2025-07-22
feat: iterator trait and improved ailist

sstadick:
Please feel free to take or leave any of this!

This adds the `find_iter` method to the `Overlapper` trait. The trick was adding the `Box<dyn>` over the return Iterator.
This also adds an iterator to ailist. 
I also added the constructor stuff I mentioned in a previous PR, reusing the decompose object. Possibly premature optimization, so that could be removed to be benchmarked at a future point.
Note that in order to return a ref iter for ailist I made the `data_list` hold the Intervals themselves. Otherwise there was nothing to return a reference to. 

TODO:

- [x] Add documentation (Ran out of time today)
- [x] Fix name of `data_list` on ailist
- [x] Remove `Default` impl and just derive it for ailist
- [x] Clean up commented out stuff

---

Added two other commits. I'd have made the PRs but Github sucks at stacked PRs from a fork it seems :( 

30d30b91c20d21efb259a064b87c0ec84c6861e8 - the `Overlapper::find_iter` method, improvements to `AiList`, impl for `find_iter` for `bits`
c7350ee7b43426dc9f61509b6363a6c304c612df - fixes bits doc tests
32a033eca442eb4dad101a540518e4442388d915 - fixes a failing test elsewhere by moving the fasta to a tempdir at the start of the tests. 
61035cb82cb834f10cd183ccf49ff6d896de5b59 - removes `Ord` trait bound which is redundant, as pointed out by @donaldcampbelljr

### DIFF
--- a/gtars/src/common/models/interval.rs
+++ b/gtars/src/common/models/interval.rs
@@ -1,17 +1,13 @@
 // https://github.com/sstadick/rust-lapper/blob/7e3904daed85181f1faa39b15f51935f13945976/src/lib.rs#L92
-use num_traits::{
-    identities::zero,
-    PrimInt, Unsigned,
-};
+use num_traits::{identities::zero, PrimInt, Unsigned};
 use std::cmp::Ordering::{self};
-
 
 /// Represent a range from [start, end)
 /// Inclusive start, exclusive of end
 #[derive(Eq, Debug, Clone)]
 pub struct Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     pub start: I,
@@ -19,10 +15,9 @@ where
     pub val: T,
 }
 
-
 impl<I, T> Ord for Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     #[inline]
@@ -35,10 +30,9 @@ where
     }
 }
 
-
 impl<I, T> Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     /// Compute the intsect between two intervals
@@ -58,7 +52,7 @@ where
 
 impl<I, T> PartialOrd for Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     #[inline]
@@ -69,7 +63,7 @@ where
 
 impl<I, T> PartialEq for Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     #[inline]
--- a/gtars/src/overlap/ailist.rs
+++ b/gtars/src/overlap/ailist.rs
@@ -9,7 +9,7 @@ use crate::common::models::Interval;
 #[derive(Debug, Clone)]
 pub struct AiList<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
     T: Eq + Clone + Send + Sync,
 {
     starts: Vec<I>,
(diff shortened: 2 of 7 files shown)


gtars/src/overlap/ailist.rs
```diff
@@ -156,25 +221,22 @@ where
...
+        data_list: &[Interval<I, T>],
     ) -> Vec<Interval<I, T>> {
         let mut results_list = Vec::new();
         let mut i = starts.partition_point(|&x| x < end);
 
         while i > 0 {
             i -= 1;
-            if start > ends[i] {
+            // maintain start inclusive, end exclusive
+            if start >= ends[i] {
```
sstadick, line 232: Note - this changes the behavior to be the same as my rust bits impl, with the half open range.

gtars/src/overlap/ailist.rs
```diff
@@ -15,17 +16,57 @@ where
...
     header_list: Vec<usize>,
-    data_list: Vec<T>,
+    data_list: Vec<Interval<I, T>>,
 }
 
-type DecomposeResult<I, T> = (Vec<I>, Vec<I>, Vec<I>, Vec<T>, Vec<Interval<I, T>>);
+#[derive(Debug)]
+struct DecomposeResult<I, T>
+where
+    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
```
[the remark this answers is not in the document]
sstadick: Good catch! removed

sstadick:
Okay, I think the things here should be in good shape overall 👍

gtars/src/common/models/interval.rs
```diff
@@ -1,28 +1,23 @@
...
 use std::cmp::Ordering::{self};
 
-
 /// Represent a range from [start, end)
 /// Inclusive start, exclusive of end
 #[derive(Eq, Debug, Clone)]
 pub struct Interval<I, T>
 where
-    I: PrimInt + Unsigned + Ord + Clone + Send + Sync,
+    I: PrimInt + Unsigned + Clone + Send + Sync,
```
nleroy917, line 10: I know @donaldcampbelljr asked about this already, but just as a follow up to that... if its a `PrimInt` (which I read as "primitive integer") shouldn't it also, by default, get `Clone`? I thought ints were a primitive type that natively supported `Copy` and by extension, Clone`

nleroy917: Hopefully that makes sense

gtars/src/overlap/ailist.rs
nleroy917: Got it -- I think its good to be consistent so this is fine

gtars/src/overlap/ailist.rs
```diff
@@ -188,20 +241,114 @@ where
...
+            inner: ailist,
+            header_list_idx: 0,
+            list_idx: None,
+            start,
+            stop,
+        }
+    }
+}
+
+impl<'a, I, T> Iterator for IterFind<'a, I, T>
```
nleroy917, line 275: Amazing thank you. I tried to do this a few weeks ago and couldn't get it. Appreacite this impl

gtars/src/overlap/ailist.rs
```diff
@@ -66,11 +121,11 @@ where
             ends,
             max_ends,
             header_list,
-            data_list,
+            stored_intervals,
         }
     }
 
-    fn find(&self, start: I, end: I) -> Vec<Interval<I,T>> {
+    fn find(&self, start: I, end: I) -> Vec<Interval<I, T>> {
```
nleroy917, line 128: Just an "engineering" thought. Not sure if it makes sense, should `find` just internally use `find_iter` that collects at the end before returning the result?

gtars/src/overlap/ailist.rs
```diff
@@ -91,29 +146,32 @@ where
...
         results_list
     }
+
+    fn find_iter<'a>(
+        &'a self,
+        start: I,
+        stop: I,
+    ) -> Box<dyn Iterator<Item = &'a Interval<I, T>> + 'a> {
+        Box::new(IterFind::new(self, start, stop))
+    }
```
nleroy917, line 161: For my own learning... `'a` is necessary as we need to annotate the `Interval` reference to say,
> This memory is valid so long as the AiList itself is valid

gtars/src/overlap/traits.rs
```diff
@@ -4,11 +4,18 @@ pub use crate::common::models::Interval;
...
+
     fn find(&self, start: I, end: I) -> Vec<Interval<I, T>>;
-}
\ No newline at end of file
+
+    fn find_iter<'a>(
+        &'a self,
+        start: I,
+        end: I,
+    ) -> Box<dyn Iterator<Item = &'a Interval<I, T>> + 'a>;
```
nleroy917, line 20: Are you able to explain why `Box` is needed here? A box'd dyn makes sense since the compiler cant possibly know which is being called at compile time since its decided at runtime (i.e. dynamic dispatch), but the Iterator is a standard interface right?

nleroy917 (approved):
This is great thank you! I think this is fine to merge, just sprinkled questions for my own learning throughout...

gtars/src/common/models/interval.rs
sstadick: That's a good question! I don't think it does need Clone. I tried removing it locally and nothing is unhappy when I take it out. I'll make a PR here shortly.

gtars/src/overlap/ailist.rs
sstadick: Yep! that's correct. "lives at lest as long as 'a" would be the way I'd read it.

PR #142, opened 2025-07-23
Add dunder functions for refget python bindings as well as type hints

donaldcampbelljr:


bindings/python/py_src/gtars/refget/__init__.pyi
```diff
@@ -0,0 +1,109 @@
...
+
+    def __repr__(self) -> str: ...
+    def __str__(self) -> str: ...
+
+class SequenceRecord:
+    """
+    A record representing a biological sequence, including its metadata and optional data.
+    """
+    metadata: SequenceMetadata
+    data: Optional[bytes] # Vec<u8> maps to bytes in Python
```
nleroy917, line 36: Confirming this one? Its both 1) sometimes `None` and then 2) the data can only be bytes?

bindings/python/py_src/gtars/refget/__init__.pyi
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Retrieves a sequence record by its SHA512t24u or MD5 digest.
+        """
+        ...
+    def get_sequence_by_collection_and_name(self, collection_digest: str, sequence_name: str) -> Optional[SequenceRecord]: ...
+    def get_substring(self, seq_digest: str, start: int, end: int) -> Optional[str]: ...
+    def write_store_to_directory(self, root_path: Union[str, PathLike], seqdata_path_template: str) -> None: ...
+
+    @classmethod
+    def load_from_directory(cls: Type["GlobalRefgetStore"], root_path: Union[str, PathLike]) -> "GlobalRefgetStore": ...
```
nleroy917, line 88: Maybe consider adding some more docstrings to these? Unless they are self-explanatory to a user.
```python
def import_fasta(self, file_path: Union[str, PathLike]) -> None: ...
    """
    Import a new fasta sequence from disk
    
    Args:
        file_path - str - the path to import from
    """
```

nleroy917 (approved):
Opportunity for more documentation in the doc strings, but it's honestly good to go

PR #145, opened 2025-07-28
Add ability to retrieve sequences from Store when using a BED file

donaldcampbelljr:
User can now retrieve sequences from a GlobalRefgetStore when providing a BED file and a sequence collection digest.
- get_seqs_bed_file -> returns retrieved sequences as a fasta file
- get_seqs_bed_file_to_vec -> Vec<RetrievedSequence>

### DIFF
--- a/gtars/src/refget/store.rs
+++ b/gtars/src/refget/store.rs
@@ -10,14 +10,17 @@ use chrono::Utc;
 use memmap2::Mmap;
 use serde::{Deserialize, Serialize};
 use std::fs::{self, File};
-use std::io::{BufRead, Write};
+use std::io::{BufRead, BufReader, Read, Write};
 use std::str;
-
+use flate2::read::GzDecoder;
 use super::encoder::decode_substring_from_bytes;
 use super::encoder::SequenceEncoder;
 use crate::common::utils::get_dynamic_reader;
 use crate::refget::fasta::read_fasta_refget_file;
-use crate::refget::hashkeyable::HashKeyable; // Import the HashKeyable trait for converting types to a 32-byte key
+use crate::refget::hashkeyable::HashKeyable;
+use crate::uniwig::reading::parse_bedlike_file;
+use crate::uniwig::utils::get_file_info;
+// Import the HashKeyable trait for converting types to a 32-byte key
 
 // Import collection types
 use super::collection::{SequenceCollection, SequenceMetadata, SequenceRecord};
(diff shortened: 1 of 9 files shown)


gtars/src/refget/store.rs
```diff
@@ -250,6 +262,253 @@ impl GlobalRefgetStore {
...
+            .open(output_file_path)?;
+
+        // Read in Bed file and get the sequences, writing to output file along the way
+        let path = Path::new(bed_file_path);
+        let file_info = get_file_info(&path.to_path_buf());
+        let is_gzipped = file_info.is_gzipped;
+
+        let opened_bed_file = File::open(path)?;
+
+        let reader: Box<dyn Read> = match is_gzipped {
```
nleroy917, line 294: Dont we have a utility function for this?

gtars/src/refget/store.rs
```diff
@@ -250,6 +262,253 @@ impl GlobalRefgetStore {
...
+
+        for (line_num, line_result) in reader.lines().enumerate() {
+            let line_string = match line_result {
+                Ok(s) => s,
+                Err(e) => {
+                    eprintln!(
+                        "Warning: Skipping line {} due to I/O error while reading: {}",
+                        line_num + 1,
+                        e
+                    );
```
nleroy917, line 313: Are we ok just printing the error? Should we panic or return `Err` here?

nleroy917: I.e. this can be condensed to:
```rust
let line_string = line_result?;
```

gtars/src/refget/store.rs
```diff
@@ -250,6 +262,253 @@ impl GlobalRefgetStore {
...
+            output_file.write_all(retrieved_substring.as_ref())?;
+        }
+
+        Ok(())
+    }
+
+    /// Given a Sequence Collection digest and a bed file path,
+    /// retrieve sequences and return them as a Vec<RetrievedSequence>.
+    /// Lines that cannot be parsed or sequences that cannot be retrieved will be skipped
+    pub fn get_seqs_bed_file_to_vec<K: AsRef<[u8]>>(
```
nleroy917, line 403: Seems like a lot of duplicated functionality. Maybe we could abstract that out and then return an iterator lover the sequence collection and/or bedfile, then one can call it and dump to file or collect into `Vec`

nleroy917 (approved):
My only qualms are with the `get_seqs_bed_file*` functions. They seem to duplicate functionality and a core function that instead returns an iterator could be useful to then create mini wrappers that consume the iterator into a vec or into a file, etc

PR #150, opened 2025-08-01
Introduce wasm bindings (and a bit more)

nleroy917:
This PR introduces web-assembly bindings for gtars so that we can use some of the features in a browser setting. This is nice for heavy processing without a server. To do this, I had to wall-off a few modules and features behind feature flags-- namely `uniwig`, `bbcache`, and anything using `reqwest`.

The reasons for each are as follows:
1. `uniwig`: pulls in `bigtools` which pulls in a compression library (`libdeflate-sys`) which wont compile to wasm since it contains some custom C code. Its possible this can be resolved in the near future, however: https://github.com/rustwasm/team/issues/291
2. `bbcache`: pulls in `biocfilecachers`, which pulls in `diesel` which was giving me a lot of troubles compiling (I think its a sqlite thing).
3. `reqwest` was also giving me a ton of issues, so I walled that off since it's only used in like one single function.
4. `hf-hub` is used to instantiate tokenizers from pre-trained models on huggingface, and that also had problems, so that's now a feature too.

I basically was able to make a ton of this stuff "opt-in" and so only when you really need that module will it compile that code for you. For example, in the python bindings, we actually **do** need the `request` features; so we enable those when we pull `gtars` in as a dependency:
```toml
# Cargo.toml
[dependencies]
gtars = { path="../../gtars", features=["http"]}
```

However, the wasm bindings don't need any of that, so we are choosing to **not** opt-in so it doesn't get compiled... and this enables us to get things working

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,5 +18,5 @@ jobs:
       run: cargo build --verbose
       working-directory: ./gtars
     - name: Run tests
-      run: cargo test --verbose
+      run: cargo test --verbose --all-features
       working-directory: ./gtars


nleroy917:
tests and builds will fail unless explicitly run with the feature:
```
cargo test --features bbcache
cargo build --features bbcache
```

nleroy917:
Aug-06-2025 09-26-20

nleroy917:
I'll probably close this in favor of #162

PR #151, opened 2025-08-07
fix bbcache

khoroshevskyi:


gtars/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gtars"
-version = "0.3.0"
+version = "0.3.1"
```
donaldcampbelljr, line 3:
```suggestion
version = "0.4.0"
```

gtars/src/main.rs
```diff
@@ -2,7 +2,7 @@ use anyhow::Result;
 use clap::Command;
 // go through the library crate to get the interfaces
 use gtars::overlap;
-use gtars::bbcache;
+// use gtars::bbcache;
```
donaldcampbelljr, line 5: these should be added back in right?

donaldcampbelljr (approved):
Needs to add the bbcache CLI and mods back into the code (they are commented out) but it looks fine otherwise.

PR #158, opened 2025-08-21
Fix accumulation during scoring, remove unnecessary additional count

donaldcampbelljr:
This PR does three things:

1. I believe I've discovered a fix for unwanted accumulation during counting while using the score column instead of the default `1` value. Essentially, we have to ensure that the most recent added score is the one that is decremented. Otherwise, the accumulation has some remainder that hangs around between start sites and before encountering the next end site. I confirmed this on some test cases and now the scored accumulation have the same shape as the non scored accumulations.

2. While looking at the counting logic, I noticed that the counts are incremented between the start site iteration and before closing out the remaining end sites. This is from the original implementation. However, we check to see if the count goes below 0 and if it does, we set it to 0 anyway. Therefore, I believe this additional count is superfluous and should be removed. You can see from the test suite examples that it does change the output slightly towards the end of the out put files.

3. adds ability to read an input file list (as .txt) for uniwig


gtars/src/uniwig/counting.rs
```diff
@@ -118,7 +120,7 @@ pub fn start_end_counts(
         prev_coordinate_value = adjusted_start_site.0;
     }
 
-    count += 1; // We must add 1 extra value here so that our calculation during the tail as we close out the end sites does not go negative.
+    //count += 1; // We must add 1 extra value here so that our calculation during the tail as we close out the end sites does not go negative.
```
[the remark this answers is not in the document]
nleroy917: I think he mentioned that this was left over from the C code and probably not necessary

gtars/src/uniwig/counting.rs
```diff
@@ -72,10 +72,10 @@ pub fn start_end_counts(
 
         adjusted_start_site = coordinate_value;
         adjusted_start_site.0 = coordinate_value.0 - smoothsize;
-
         let current_score = adjusted_start_site.1;
 
         count += current_score;
+        collected_counts.insert(0,current_score); // needed to later decrement in the correct order
```
nleroy917, line 78: `cargo fmt`

gtars/src/uniwig/counting.rs
```diff
@@ -176,6 +178,7 @@ pub fn core_counts(
     let mut current_end_site: (i32, i32);
 
     let mut collected_end_sites: Vec<(i32, i32)> = Vec::new();
+    let mut collected_counts: Vec<i32> = Vec::new();
```
nleroy917, line 181: Does this type need to be annotated? I feel like I remember Seth arguing that we should let Rust do the type inference

nleroy917 (approved):
Looks good!

donaldcampbelljr:
I also confirmed that this does not break the workflow for creating wiggles and then using wigToBigWig to convert to bw. We are not really doing that workflow any longer (I believe) but wanted to ensure it still worked.

PR #161, opened 2025-08-29
wasm bindings for refget

donaldcampbelljr:
Attempt wasm bindings for GlobalRefgetStore

### DIFF
--- a/bindings/wasm/src/lib.rs
+++ b/bindings/wasm/src/lib.rs
@@ -2,6 +2,8 @@ mod digests;
 mod tokenizers;
 mod utils;
 
+use std::str::EncodeUtf16;
+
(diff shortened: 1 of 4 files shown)


donaldcampbelljr:
## WIP

Right now the file loader will only accept a .fa file, if it is gzipped it will give an error:
```
    FASTA parse error: expected '>' but found '\u{1f}' at file start, line 1.
```

Also, if you give it a large reference genome (gigabytes), it will fail:

```
Error Processing File
Error: The requested file could not be read, typically due to permission problems that have occurred after a reference to a file was acquired.

File: hg38.p14.fa
```

Using a much smaller file (300mb) appears to work.

PR #162, opened 2025-09-03
Convert gtars into a workspace

nleroy917:
I spent some time making `gtars` a workspace. It actually wasn't that bad at all, and I quite like it. It allows us to get much more granular about dependencies, features, etc. For example, we can pin ultra-common dependencies like `anyhow`, `serde` at the workspace level, but then dive into crates and then pull more things in as needed (`bigtools`, `noodles`, etc).


nleroy917:
@khoroshevskyi will have to reimplement #151 and #159 in the workspace after we merge this

PR #165, opened 2025-09-09
Bed reference validation implemented in Rust

khoroshevskyi:
implementation attempt of bed reference validator in rust


khoroshevskyi:
After testing the new BED reference validator in Rust, we managed to reduce the runtime from 0.05s to 0.04s per file. With a little over 200K files to process, an additional 0.01s improvement would translate to about 30 minutes saved in a 24-hour processing run.

While this shows the Rust implementation is efficient, it is also harder to maintain compared to the existing implementation in BEDboss (Python). After discussion, we decided to close this issue and keep the main implementation of the BED reference validator in BEDboss (Python).

PR #168, opened 2025-09-10
Changed vec to vecdeque to improve insertion and deletion from the vec

khoroshevskyi:


khoroshevskyi:
Because of the merge conflicts, this code will be added to gtars with workspace refactoring

Issue #148, opened 2025-07-30
Convert gtars to a workspace


nleroy917:
I know we've discussed this a lot already, but I’d like to formally make the case for converting `gtars` into a workspace. The crate has become too bloated, and it’s making it increasingly difficult to build for multiple targets. Specifically, the Python bindings, R bindings, and CLI are all now transitively dependent on code they don’t even use, and it’s getting frustrating.

## Example 1: Python bindings broken by transitive dependency

With the release of `v0.3.0`, we couldn’t build the Python bindings due to this linker error:
```
cannot find -lsqlite3.lib
```

This comes from `diesel`, which is pulled in by `bioccachers`, which is pulled in by `bbcache`– a module inside `gtars`. Even though `bbcache` isn’t used at all in the Python bindings, it still gets compiled and brings its dependencies along with it.

If `gtars` were a workspace, we could isolate core modules like `tokenizers`, `refget`, and `RegionSet` into independent crates and import only what we need. This would eliminate the `sqlite3` dependency for the Python bindings entirely.

## Example 2: WebAssembly blocked by openssl

I attempted to build WebAssembly (`wasm`) bindings for `gtars`, but ran into build failures due to an `openssl` dependency. This comes from `bbcache` and `RegionSet`, even though I only needed `refget` and the overlap logic.

Again, with a workspace structure, I could just depend on the minimal subset required, avoiding the `openssl` problem entirely.

donaldcampbelljr:
You've written some of the cons of the current approach, but what are the cons of making this a workspace?

nleroy917:
One immediate con that I can think of is that if we intend to put things on `crates.io` they must all be separated and published separately

nleroy917:
> Using no emojis and minimal em-dashes, can you please provide some pros and cons to building a rust codebase using Cargo workspaces versus namespaced modules? Please output in raw markdown to paste into a github issue

## Pros and Cons: Cargo Workspaces vs. Namespaced Modules in Rust

### Cargo Workspaces

**Pros**
- **Modular compilation:** Each crate in the workspace is compiled independently, enabling incremental compilation and caching across crates.
- **Logical separation:** Ideal for organizing code with clear boundaries (e.g., CLI, core logic, bindings, plugins).
- **Testing isolation:** Each crate can have its own test suite and `Cargo.toml`, making unit testing and CI workflows cleaner.
- **Fine-grained dependency control:** Different crates can depend on different versions of the same external libraries, or expose different feature flags.
[...]

PR #169, opened 2025-09-10
gtars v0.5.0 release (as workspace)

nleroy917:
This will be the final merge to convert into a workspace, addressing #148 

I realized that the `docs/` folder got dropped. That has our logos, so we should add that back in too

Before merging, I'd emphasize those with ownership over crates (e.g. Nathan L w/ tokenizers, Donald with uniwig, Alex with RegionSets) to just do some cursory testing to make sure things seem correct. All of our unit tests are passing, but it's still a good gut check.

### TODO:
- [x] local, gut-check testing
- [x] add `docs` back in
- [x] double-check versioning across crates
---
- [ ] Add docs how to install each crate and how to install cli
- [ ] Update versions of each crate
- [ ] Update bioconda workflow (TODO fater release)
- [ ] Release as alpha version ?

### DIFF
--- a/gtars-cli/src/bbcache/handlers.rs
+++ b/gtars-cli/src/bbcache/handlers.rs
@@ -1,9 +1,9 @@
 use std::fs::read_dir;
 use std::path::PathBuf;
 
+use clap::ArgMatches;
 use gtars_bbcache::client::BBClient;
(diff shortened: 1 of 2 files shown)


tokenizers/src/cli.rs
donaldcampbelljr, line 1: This entire file seems commented out. Please remove it if we do not need it.

PR #174, opened 2025-10-01
Introduce wasm bindings (finally)

nleroy917:
This isn't complete. but the new workspace makes it a lot more amenable to create wasm bindings. To start, I had to add two new feature-flags to `gtars-core`:
1. `bigbed` (walls off `bigtools` which wont compile to wasm), and
2. `http` (walls off `reqwest` which also wouldn't compile to wasm)

@khoroshevskyi these are both `RegionSet` implementation features so maybe we can discuss that a bit more. But it was actually quite easy to do.


nleroy917:
Close to done. Just need to test it out and then of course tackle publishing -- ideally in CI/CD

nleroy917:
@donaldcampbelljr @khoroshevskyi still working on this I think... but its close

PR #179, opened 2025-10-14
Refactor IGD, Uniwig, bbcache tests

### LINKED ISSUE #178: Refactor tests
We have a test file that contains tests from multiple crates. The rust approach is to place these tests within the module/library where the primary code is executing. Some crates are already doing this, but **IGD and Uniwig** are not.

donaldcampbelljr:
Fixes #178

### DIFF
--- a/gtars-bbcache/Cargo.toml
+++ b/gtars-bbcache/Cargo.toml
@@ -18,7 +18,7 @@ anyhow = { workspace = true }
 serde_json = { workspace = true }
 
 # our code
-gtars-core = { path = "../gtars-core", version="0.5.0" }
+gtars-core = { path = "../gtars-core", version="0.5.1" }
 
 [dev-dependencies]
 rstest = "0.26.1"


nleroy917:
Hey thanks for tackling this. My tests are failing l locally though? Seems like the we are missing the core crate as a dependency in the `gtars-bbcache` module

gtars-bbcache/src/lib.rs
```diff
@@ -1,3 +1,86 @@
...
+    }
+
+    #[fixture]
+    fn bbid() -> PathBuf {
+        "6b2e163a1d4319d99bd465c6c78a9741".into()
+    }
+
+    #[fixture]
+    fn path_to_bedset() -> PathBuf {
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
nleroy917: This seems useful! Is this like instead of doing `../tests/data`? If so, we should mirror this to other tests

gtars-igd/Cargo.toml
```diff
@@ -7,9 +7,15 @@ license = "MIT"
...
+[dev-dependencies]
+rstest = "0.26.1"
+tempfile = "3.23.0"
+
+
 # shared workspace dependencies
 anyhow = { workspace = true }
-byteorder = { workspace = true }
\ No newline at end of file
+byteorder = { workspace = true }
```
nleroy917: Two byteorder deps?

gtars-igd/src/lib.rs
```diff
@@ -8,3 +8,370 @@ pub mod consts {
...
+    fn test_igd_create_short_long_regions() {
+        // Depending on start and end coordinates which are divided by nbp=16384
+        // the number of tiles per ctg are adjusted, this tests to ensure they are created appropriately
+        let tempdir = tempfile::tempdir().unwrap();
+        let path = PathBuf::from(&tempdir.path());
+        let db_path_unwrapped = path.into_os_string().into_string().unwrap();
+        let db_output_path = db_path_unwrapped;
+
+        let path_to_crate = env!("CARGO_MANIFEST_DIR");
+        let testfilelists = format!("{}{}", path_to_crate, "/../tests/data/igd_file_list_01/");
```
nleroy917: `/../` is weird to me... whats that?

nleroy917 (changes requested):
I think we just need to fix that core issue with the dev dependencies in the `gtars-bbcache` `Cargo.toml` file.

Also I'm curious about the use of `/../` and `CARGO_MANIFEST_DIR`. Mostly for my own learning. Again thanks for tackling this!

gtars-bbcache/src/lib.rs
donaldcampbelljr: Yes, it is better for cross platform compatibility. Original way may give issues on non unix platforms.

gtars-bbcache/src/lib.rs
```diff
@@ -1,3 +1,86 @@
...
+            .parent()
+            .unwrap()
+            .join("tests/data/bedset")
+    }
+
+    use super::client::BBClient;
```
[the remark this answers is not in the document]
donaldcampbelljr: done

PR #185, opened 2025-10-20
feat: Create core and binding functionality for basic region-set statistics + extras

khoroshevskyi:
Lots of things in here (nathan):
- #187
- #192
- #188
- #193 

## Changes:
- Added stats calculation in RegionSet struct
- Added wasm bindings for Regionset

## TODO:
- [x] Add docstrings for functions
- [x] Change struct item names that are mentioned in PR comments

### DIFF
--- a/gtars-genomicdist/src/models.rs
+++ b/gtars-genomicdist/src/models.rs
@@ -33,5 +33,5 @@ pub struct RegionBin {
     /// End position of the bin
     pub end: u32,
     /// Number of regions overlapping this bin
-    pub n: u32
+    pub n: u32,
 }
--- a/gtars-genomicdist/src/statistics.rs
+++ b/gtars-genomicdist/src/statistics.rs
@@ -114,7 +114,7 @@ impl GenomicIntervalSetStatistics for RegionSet {
                         chr: k.chr.clone(),
                         start: k.start,
                         end: k.end,
-                        n: 1
+                        n: 1,
                     },
                 );
             }
@@ -153,6 +153,4 @@ mod tests {
         assert_eq!(chr1_stats.start_nucleotide_position, 5);
         assert_eq!(chr1_stats.end_nucleotide_position, 36);
     }
-
-    
 }


gtars-core/src/models/region.rs
```diff
@@ -41,3 +41,22 @@ impl Display for Region {
...
+#[derive(Debug, Clone)]
+pub struct ChromosomeStats {
+    pub chromosome: String,
+    pub count: u32, // number of regions
+    pub start: u32,
+    pub end: u32,
+    pub minimum: u32,
+    pub maximum: u32,
+    pub mean: f64,
+    pub median: f64,
```
nleroy917: Can we come up with some more descriptive names for these maybe? Right now its not immediately clear to me what these are.

gtars-core/src/models/region_set.rs
```diff
@@ -531,6 +532,56 @@ impl RegionSet {
         }
         total_count
     }
+
+    ///
+    /// Calculate statistics
+    ///
+    pub fn calculate_statistics(&self) -> HashMap<String, ChromosomeStats> {
```
nleroy917: This returns chrom stats per chrom, right? Should we maybe make the function + doc string reflect that? Maybe:
```rust
fn calculate_chromosome_statistics
```

gtars-core/src/models/region_set.rs
```diff
@@ -531,6 +532,56 @@ impl RegionSet {
...
+                    as f64
+                    / 2.0
+            } else {
+                sorted_widths[(count / 2) as usize] as f64
+            };
+
+            stats.insert(
+                chr.clone(),
+                ChromosomeStats {
+                    chromosome: chr.clone(),
```
nleroy917: This field feels redundant now that I am looking at it.. the key already indicates which Chrom we are on. maybe its good to keep that, but it would save another clone if we got rid of this.

nleroy917 (changes requested):
Biggest thing for me is the API naming. I think it would be helpful to at least discuss two things:
1. More descriptive function/attribute names, and
2. Where to actually put this `ChromosomeStats` module in the context of the library/project. I might even suggest an entire crate dedicated to qc/stats

More abstract and thinking bigger... it could be useful to pull this stuff out into functions.
```rust
fn compute_chromosome_distribution(rs: &RegionSet): ...
fn compute_neighbor_distnaces(rs: &RegionSet): ...
```
etc etc etc. Maybe like genomicDistributions but in rust? Just make it its own crate? maybe Im mostly arguing for **not** tying all this functionality to the RegionSet object itself. Feels really OOP and not very rusty, in my opinion. Very opn, however, to others thoughts.

---

Better yet, make it a trait: `ChromosomeStatistics` and then that way we can implement the trait for our types: `BedSet`, `RegionSet`, `FragmentsFile`, etc. That might be the way to go, honestly

gtars-core/src/models/region_set.rs
khoroshevskyi: it's ok for now

gtars-gd/src/statistics.rs
```diff
@@ -0,0 +1,112 @@
...
+                region_bin.n += 1;
+            } else {
+                plot_results.insert(
+                    k.digest().clone(),
+                    RegionBin {
+                        chr: k.chr.clone(),
+                        start: k.start,
+                        end: k.end,
+                        n: 1,
+                        rid: k.start / region_length,
```
nsheff: it's not clear what this line is. what is `rid` ?

gtars-overlaprs/src/genome_index.rs
```diff
@@ -0,0 +1,678 @@
...
+//!
+//! ## Basic Usage
+//!
+//! ```
+//! use gtars_overlaprs::{OverlapperType, genome_index::IntoGenomeIndex};
+//! use gtars_core::models::{Region, RegionSet};
+//!
+//! // Create a genome-wide set of intervals (e.g., gene annotations)
+//! let genes = vec![
+//!     Region { chr: "chr1".to_string(), start: 1000, end: 2000, rest: Some("BRCA1".to_string()) },
```
nsheff: I think using a Region object like this is really atypical. It's also going to be slower.

It feels better to have your examples use vectors of regions, rather than individual region elements.

gtars-overlaprs/src/genome_index.rs
```diff
@@ -0,0 +1,678 @@
+//! Genome-wide interval indexing for efficient multi-chromosome overlap queries.
```
nsheff: It's not clear to me why this data structure is called "Genome Index".

What does the Index mean exactly? It's not really explained.

gtars-overlaprs/src/genome_index.rs
```diff
@@ -0,0 +1,678 @@
...
+    RegionParsingError(String),
+    /// Error converting interval coordinates to the required integer type.
+    #[error("Error converting interval coordinates to u32: start={0}, end={1}")]
+    CoordinateConversionError(String, String),
+}
+
+/// A genome-wide index for efficient overlap queries across multiple chromosomes.
+///
+/// `GenomeIndex` maintains a separate overlap data structure (either [`AiList`] or [`Bits`])
+/// for each chromosome, enabling efficient queries across the entire genome. This is particularly
```
nsheff: The last sentence here doesn't really make sense: 

finding overlaps between genomic features on different chromosomes is impossible by definition.

gtars-gd/src/statistics.rs
```diff
@@ -0,0 +1,112 @@
...
+            let widths: Vec<u32> = regions.iter().map(|r| r.width()).collect();
+            let minimum = *widths.iter().min().unwrap_or(&0);
+            let maximum = *widths.iter().max().unwrap_or(&0);
+
+            let earliest_position = regions.iter().map(|r| r.start).min().unwrap_or(0);
+            let end_position = regions.iter().map(|r| r.end).max().unwrap_or(0);
+
+            let mean = widths.iter().sum::<u32>() as f64 / count as f64;
+
+            let mut sorted_widths = widths.clone();
```
donaldcampbelljr: Is there a way for us to do this without using a clone? Can we attempt to take a slice and do the same thing?

gtars-gd/src/statistics.rs
```diff
@@ -0,0 +1,112 @@
...
+use gtars_overlaprs::genome_index::IntoGenomeIndex;
+use gtars_overlaprs::OverlapperType;
+use std::collections::HashMap;
+
+pub trait Statistics {
+    /// Calculate basic statistic for each region in the chromosome
+    fn calculate_chr_statistics(&self) -> HashMap<String, ChromosomeStats>;
+
+    /// Generate chrom distribution data based on regions used in the RegionSet
+    fn generate_region_distribution(&self, n_bins: u32) -> HashMap<String, RegionBin>;
```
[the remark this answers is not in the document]
nleroy917: So maybe to be more specific:
```rust
fn generate_region_distribution_with_bin_size(&self, n_bins: u32) {
  // pass
}

fn generate_region_distribution(&self) {
   self.generate_region_distribution_with_bin_size(&self, 10)
}
```

gtars-gd/src/statistics.rs
nleroy917: Yeah I dont think you need to clone here... nothing above should be taking ownership of that data, right?

gtars-overlaprs/src/multi_chrom_overlapper.rs
```diff
@@ -0,0 +1,678 @@
+//! Genome-wide interval indexing for efficient multi-chromosome overlap queries.
```
nleroy917: We have three structs:
1. `AIList`
2. `Bits`
3. `GenomeIndex`

We also have two traits:
1. `Overlapper`
2. `IntoGenomeIndex`

`AIList` and `Bits` implement the `Overlapper` trait. It allows them to perform overlap queries on a single chromosome (a set of intervals). `GenomeIndex` expands this and makes them more useful for genome-wide overlap queries. Its simply a wrapper around the `Overlapper` trait:
```rust
struct GenomeIndex {
   inner: HashMap<String, Box<dyn Overlapper>>
}
```
I'm going to just rename it to be `MultiChromOverlapper` and then create a new trait called `IntoMultiChromOverlapper`. I think that i see the simplicity and niceness of `Overlapper` and `MultiChromOverlapper` traits, but there will only be one implementation of `MultiChromOverlapper` trait (the previously named `GenomeIndex`).

So to that end.. I don't really see the need to make the names match. One is just a more useful abstraction of the other?   

Anyways just thinking out loud -- I'm fine to rename away from `GenomeIndex` but I might stop there and not try to create parallels between trait names

gtars-overlaprs/src/ailist.rs
```diff
@@ -5,7 +5,37 @@ use num_traits::{PrimInt, Unsigned};
...
+/// // Query for genes overlapping position 1800-2200
+/// let overlaps = ailist.find(1800, 2200);
+/// assert_eq!(overlaps.len(), 2); // GENE1 and GENE2
+///
+/// // Check if the list is empty
+/// assert!(!ailist.is_empty());
+/// assert_eq!(ailist.len(), 3);
+/// ```
 #[derive(Debug, Clone)]
 pub struct AiList<I, T>
```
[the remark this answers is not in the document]
nleroy917: Fixed!

PR #186, opened 2025-10-20
Add bloom filter as optional dependency to IGD

donaldcampbelljr:
Wanted to add the bloom filter work back into dev after the major refactoring we did over the last few months.

However, I believe both the external bloom API changed as well as our own Region and Tokenizer API. so this has proved more challenging that originally planned.

### DIFF
--- a/gtars-igd/src/igdbloom.rs
+++ b/gtars-igd/src/igdbloom.rs
@@ -96,14 +96,17 @@ where
     }
 }
 
-pub fn load_bloom_filter_from_disk(
-    load_path: &str,
-) -> Result<Bloom<String>, Box<dyn std::error::Error>> {
-    let bytes = fs::read(load_path)?;
+pub fn load_bloom_filter_from_disk<P>(
+    load_path: P,
+) -> Result<Bloom<String>, Box<dyn std::error::Error>>
+where
+    P: AsRef<Path>,
+{
+    let bytes = fs::read(&load_path)?;
 
     let filter = Bloom::from_bytes(bytes).map_err(|e| format!("Bloom filter error: {}", e))?;
 
-    println!("Successfully loaded bloom filter from: {}", load_path);
+    println!("Successfully loaded bloom filter from: {}", load_path.as_ref().display());
     Ok(filter)
 }
 
@@ -164,15 +167,18 @@ pub fn process_bed_directory(
     Ok(processed_count)
 }
 
-pub fn load_bloom_directory(
-    bloom_directory: &str,
-) -> Result<HashMap<String, Bloom<String>>, Box<dyn std::error::Error>> {
-    let bloom_path = Path::new(bloom_directory);
+pub fn load_bloom_directory<P>(
+    bloom_directory: P,
+) -> Result<HashMap<String, Bloom<String>>, Box<dyn std::error::Error>>
+where
+    P: AsRef<Path>,
+{
+    let bloom_path = bloom_directory.as_ref();
 
     if !bloom_path.exists() || !bloom_path.is_dir() {
         return Err(format!(
             "Bloom directory does not exist or is not a directory: {}",
-            bloom_directory
+            bloom_directory.as_ref().display()
         )
         .into());
     }
@@ -213,7 +219,7 @@ pub fn load_bloom_directory(
     println!(
         "Loaded {} bloom filters from directory: {}",
         bloom_filters.len(),
-        bloom_directory
+        bloom_directory.as_ref().display()
     );
     Ok(bloom_filters)
 }


donaldcampbelljr:
I now have a basic example working for crafting a bloom filter based on a tokenized bed file, exporting it to disk, re-loading it from disk, and running a query based on a single region.

This query would only return true if there is an _exact_ match. It does not account for overlaps.

gtars-igd/Cargo.toml
```diff
@@ -10,11 +10,17 @@ repository = "https://github.com/databio/gtars"
 byteorder.workspace = true
 # our code
 gtars-core = { path = "../gtars-core", version="0.5.0" }
+gtars-tokenizers = { path = "../gtars-tokenizers", version="0.5.1", optional = true }
```
nleroy917: just make sure that the version is correct and up to date!

gtars-igd/Cargo.toml
```diff
@@ -10,11 +10,17 @@ repository = "https://github.com/databio/gtars"
...
 tempfile = "3.23.0"
-
+pretty_assertions = "1.4.1"
 
 # shared workspace dependencies
 anyhow = { workspace = true }
+
+[features]
+# Feature 'bloom' links to the dependency key 'bloom' (which is the correct crate name).
+bloom = ["dep:bloomfilter", "dep:gtars-tokenizers"]
```
nleroy917: Great 👍🏼👍🏼👍🏼

gtars-igd/src/igdbloom.rs
```diff
@@ -0,0 +1,319 @@
...
+                Err(e)
+            }
+        }
+    } else {
+        println!("Parent directory already exists: {}", parent_directory);
+        Ok(())
+    }
+}
+
+#[cfg(feature = "bloom")]
```
nleroy917: Is it worth just doing a module-level feature flag to gate this? so in the `lib.rs`:
```rust
#[cfg(feature="bloom")]
pub mod igdbloom.rs
```

gtars-igd/src/lib.rs
```diff
@@ -3,6 +3,9 @@
 pub mod create;
 pub mod search;
 
+#[cfg(feature = "bloom")]
```
nleroy917: Since this feature flag is here... you might not need them in `igdbloom.rs`, right?

nleroy917 (approved):
Looks good! my only two thougths are:

1. See if you can use `MultiChromOverlapper` instead of `Tokenizer` (its much more bare-bones and simple), and then
2. see if its possible to remove those individual feature flags in on each function inside `igdbloom.rs`, instead opting to just gate the whole module entirely

gtars-igd/src/igdbloom.rs
```diff
@@ -0,0 +1,319 @@
...
+                    println!("Loading bloom filter: {}", bloom_file);
+
+                    match load_bloom_filter_from_disk(bloom_file) {
+                        Ok(filter) => {
+                            bloom_filters.insert(filename, filter);
+                        }
+                        Err(e) => {
+                            eprintln!("Failed to load bloom filter from {}: {}", bloom_file, e);
+                            // Continue loading other files even if one fails
+                        }
```
khoroshevskyi: Do we want to collect this information somewhere, and create log file out of it? Just thoughts

gtars-igd/src/igdbloom.rs
donaldcampbelljr: Yeah I just confirmed that this works. Originally, it was giving me trouble, so I put the feature flag everywhere but it now seems unnecessary so I've removed all but the module level flag.

gtars-igd/src/README.md
```diff
@@ -22,4 +22,88 @@ cargo run igd create --output /home/drc/IGD_TEST_2/igd_rust_output/ --filelist /
...
+    // Directory where .bloom files will be saved
+    let output_dir = "/path/to/bloom/output";
+    
+    // Bloom filter parameters
+    let num_of_items = 10000;           // Expected number of tokens per file
+    let false_positive_rate = 0.01;      // 1% false positive rate
+    
+    // Create tokenizer from a sample BED file
+    let sample_bed = "/path/to/sample.bed"; // Preferably this should be a UNIVERSE BED file
+    let tokenizer = Tokenizer::from_auto(sample_bed)?;
```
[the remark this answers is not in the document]
nleroy917: The tokenizer and its API are sort of engineered for machine learning applications, so there might be some headaches or "surprises" when you use it for things that aren't machine learning.

Two examples I could think of...

1. If there are no overlaps for a given region, I believe it spits out `chrUNK:0-0`. Not sure if that's the intended behavior of IGD. It might not matter... but it could be a weird side effect for your use-case.
2. Its vocab-size is tied to special tokens. If you give it a bed file with 10 regions, the vocab size is going to be 17 (it auto-adds seven special tokens). Again, this might not matter, but I could see it tripping you up and being a hard-to-debug thing in the future should it cause a problem.

But mostly, they are just meant for ML applications... You _might_ be better off using the overlappers, since the API surface area is smaller and the internals are probably much closer to what your mental model expects.

All that being said, it probably doesn't matter for a first iteration. The overlappers are meant to be "primitives" from which we can build things, and I see IGD + tokenizers as downstream of that... right now, IGD is downstream of the tokenizers, which is downstream of the overlappers.

gtars-igd/src/igdbloom.rs
```diff
@@ -0,0 +1,311 @@
...
+        println!("Parent directory already exists: {}", parent_directory);
+        Ok(())
+    }
+}
+
+pub fn write_bloom_filter_to_disk(
+    igd_bloom_filter: Bloom<String>,
+    save_path: String,
+) -> Result<(), std::io::Error> {
+    let bytes = igd_bloom_filter.to_bytes();
```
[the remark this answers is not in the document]
khoroshevskyi: would it be better to do something as: AsRef() ?

PR #190, opened 2025-10-23
feat, chores, docs: Better overlapprs docs, new genome index functionality, etc

nleroy917:
Per the request of @khoroshevskyi, I am adding better documentation for using the `overlaprs` crate


nleroy917:
Ok -- this should be super thorough now

Issue #198, opened 2025-10-25
Improve the `scoring` API


nleroy917:
The scoring module was sort of half baked and not a lot of thought went into the API. It would be good to do another pass at this to improve the API/functionality

PR #201, opened 2025-10-25
add --barcode mode to scoring function

nsheff:
adds a matrix market output format for scoring large matrices

### DIFF
--- a/gtars-cli/src/scoring/handlers.rs
+++ b/gtars-cli/src/scoring/handlers.rs
@@ -5,39 +5,73 @@ use anyhow::Result;
 use clap::ArgMatches;
 
 use gtars_scoring::consts;
-use gtars_scoring::{ConsensusSet, FragmentFileGlob, ScoringMode, region_scoring_from_fragments};
+use gtars_scoring::{ConsensusSet, FragmentFileGlob, ScoringMode, region_scoring_from_fragments, barcode_scoring_from_fragments, write_sparse_counts_to_mtx};
 
 pub fn run_scoring(matches: &ArgMatches) -> Result<()> {
-    // get arguments from CLI
-    let fragments = matches
-        .get_one::<String>("fragments")
-        .expect("A path to fragment files is required.");
+    // Check if barcode mode is enabled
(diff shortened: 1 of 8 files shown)


nsheff:
I can confirm that this works for me to create cell-by-peak matrix market files for scanpy analysis.

gtars-scoring/src/fragment_scoring.rs
```diff
@@ -117,6 +119,41 @@ pub fn region_scoring_from_fragments(
...
 
+/// Score fragments by barcode for single-cell analysis
+///
+/// Returns sparse HashMap: barcode → (peak_index → count)
+pub fn barcode_scoring_from_fragments<P: AsRef<Path>>(
+    fragment_file: P,
+    consensus: &ConsensusSet,
+) -> Result<HashMap<String, HashMap<usize, u32>>> {
+    let reader = get_dynamic_reader(fragment_file.as_ref())?;
+    let mut barcode_counts: HashMap<String, HashMap<usize, u32>> = HashMap::new();
```
nleroy917: Remove type annotation? isnt it inferred at some point?

nleroy917:
We can address streamlining and I/O in a future PR, as mentioned in #198

gtars-cli/src/scoring/handlers.rs
```diff
@@ -5,39 +5,73 @@ use anyhow::Result;
...
+            .get_one::<String>("fragments")
+            .expect("A path to fragment files is required.");
+
+        let consensus = matches
+            .get_one::<String>("consensus")
+            .expect("A path to a mapping file is required.");
+
+        let default_out = consts::DEFAULT_OUT.to_string();
+        let output = matches.get_one::<String>("output").unwrap_or(&default_out);
+        let mode = match matches.get_one::<String>("mode") {
```
donaldcampbelljr: Should `barcode` be a potential value for `mode` ? That way we do not need a new flag?

nleroy917: Yeah I agree

PR #202, opened 2025-10-25
Fix test_import_fasta to use temporary directory

### LINKED ISSUE #199: Lingering files during tests
I noticed that there are two files consistently lingering during tests:
1. `gtars-refget/tests/store_test/index.json`
2. `gtars-refget/tests/store_test/sequences.farg`

Its kind of annoying to run tests after changes then have my git tell me I have changes to commit again. can we just put these in a gitignore?

nsheff:
Resolves #199

The test_import_fasta test was writing store artifacts to a hardcoded path "tests/store_test/" without cleanup, causing lingering files to appear in git status after test runs. This fix modifies the test to use the temporary directory already created in the test, consistent with all other file-writing tests in the suite.

Changes:
- Modified store.write_store_to_directory() call to use temp_path
- Removed existing tests/store_test/ directory and all artifacts
- Test now auto-cleans when temp_dir is dropped at test completion

This follows the cleanup pattern used by test_disk_persistence and other tests that write files.


donaldcampbelljr (approved):
Looks good. Thanks for doing this. It was also annoying me earlier this week as well.

Issue #183, opened 2025-10-20
Refactor error handling.


donaldcampbelljr:
Related to comments in #182 . 

I added some `expect`methods to one of the writing functions in place of a generic `unwrap`. However, it would be nice to have better error propagation within gtars overall.

PR #208, opened 2025-10-28
refactor: move the IO functionality for regionset

nleroy917:
I basically just created two new traits in `gtars-io`:
1. `BedWrite`
2. `BigBedWrite`

And then I implemented them for `RegionSet`. This way... that functionality/business logic no longer lives inside the `RegionSet` struct, and instead it lives in the `gtars-io` module and we implement that functionality for any types we want. More idiomatic and flexible and spreads out the code.


gtars-io/src/bigbed.rs
```diff
@@ -0,0 +1,131 @@
...
+impl BigBedWrite for RegionSet {
+    fn write_bigbed<P: AsRef<Path>>(
+        &self,
+        out_path: P,
+        chrom_size: P,
+    ) -> Result<(), std::io::Error> {
+        let out_path = out_path.as_ref();
+        if out_path.exists() {
+            println!("Bed file already exists. Overwriting existing file")
+        }
```
donaldcampbelljr: We are warning the user here but not for Bed writing nor BedGZ writing. Would recommend making them consistent.

Also this should say "BigBed exists...etc"

gtars-io/src/bigbed.rs
```diff
@@ -0,0 +1,131 @@
...
+                eprintln!(
+                    "Warning:: Chromosome is not found in Chrom sizes file. Chr: '{}'",
+                    i.chr
+                );
+                if warnings_count > 40 {
+                    panic!("Incorrect chrom sizes provided. Unable to create bigBed file!");
+                }
+
+                warnings_count += 1;
+                return None;
```
donaldcampbelljr: I know we flatten below but could just `continue` instead of returning `None`?

gtars-io/src/bigbed.rs
```diff
@@ -0,0 +1,131 @@
...
+        let mut bb_out = BigBedWriter::create_file(out_path, chrom_sizes.clone())
+            .expect("Failed to create bigBed file.");
+
+        bb_out.options.max_zooms = 8;
+
+        let data = BedParserStreamingIterator::wrap_iter(region_vector.into_iter(), true);
+        bb_out
+            .write(data, runtime)
+            .map_err(|e| std::io::Error::other(e.to_string()))?;
+        Ok(())
```
donaldcampbelljr: Ok, so we create an initial file and can warn the user if that initial creation fails but what about the next streaming step? What happens if streaming fails? How is the user alerted?

gtars-io/src/bigbed.rs
```diff
@@ -0,0 +1,131 @@
...
+                        .rest
+                        .as_deref()
+                        .map_or(String::new(), |s| format!("\t{}", s)),
+                },
+            )))
+        });
+
+        #[allow(clippy::option_filter_map)]
+        // I like this because its more readable and clear whats going on
+        let region_vector = region_vector.filter(|e| e.is_some()).map(|e| e.unwrap());
```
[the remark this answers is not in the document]
nleroy917: yes -- I think I kept it there (you can see me suppress the clippy warning) since it seems more clear what's going on. But happy to swap

gtars-io/src/gtok.rs
```diff
@@ -67,10 +66,10 @@ impl<W: Write> GTokWriter<W> {
         // Header writing could be moved to it's own method
         writer
             .write_all(GTOK_HEADER)
-            .with_context(|| "Failed to write GTOK header to file!")?;
+            .map_err(|_| GtokError::HeaderWrite)?;
```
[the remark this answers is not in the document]
nleroy917: Agreed. This was me trying to work on #183.

My googling and thread reading has me semi-convinced that `anyhow` is not a good use case for us. We should be defining our own error-types

gtars-io/src/bigbed.rs
nleroy917: @khoroshevskyi ? I think that this is your code

donaldcampbelljr:
I think we are on the cusp of a situation where this would be helpful. And you've already done the hard part. Another good thing is that it sets a good example within our own code base for how to implement something like this in the after refactor, i.e. it is nice to see this in our own code rather than to go google for a more basic "hello world" example when expanding functionality in the future.

gtars-io/src/bigbed.rs
khoroshevskyi: I think this is the solution we came up. I think it's because we are using iter().map()...

gtars-io/src/bigbed.rs
```diff
@@ -0,0 +1,131 @@
...
+        let mut warnings_count: i32 = 0;
+        let region_vector = self.regions.iter().map(|i| {
+            // This if it is removing regions that are not in chrom sizes file.
+            if !chrom_sizes.contains_key(&i.chr) {
+                eprintln!(
+                    "Warning:: Chromosome is not found in Chrom sizes file. Chr: '{}'",
+                    i.chr
+                );
+                if warnings_count > 40 {
+                    panic!("Incorrect chrom sizes provided. Unable to create bigBed file!");
```
khoroshevskyi: If we are doing this restructuring, can we try to return error here instead of panic?

khoroshevskyi (approved):
As I said, this restructuring seems to be quite complicated for someone who will use `RegionSet` for the first time. But If you are saying that this is the correct way of doing, then let's do it. 
I have one comment about changing panic to error, but otherwise, looks good!

PR #214, opened 2025-12-05
Adds remote-access refgetstore and fai index capability

nsheff:


khoroshevskyi (approved):
After reviewing the code and the documentation at a high level, everything looks good to me

PR #243, opened 2026-03-09
try implementing a gtars-sc crate

sanghoonio:
This is an attempt to add a single-cell analysis framework like seurat/scanpy in gtars, with a cli that focuses on AI-integrated workflows and wasm bindings for end to end in-browser analysis


sanghoonio:
Off the top of my head, just wasm support so I can try building an in-browser end to end analysis tool. I haven't actually explored snapatac2 very much.

PR #244, opened 2026-03-11
Fix genomicdist bugs

nsheff:
- Fix closest() nearest-neighbor search to use bidirectional expanding scan instead of fixed ±2 window that missed true nearest neighbors (5 new tests)
- Replace unsafe native-endian f64 serialization in signal.rs with safe explicit little-endian encoding, bump SIGM_VERSION to 2 (3 new tests)
- Fix WASM classify_bed_js to return Result instead of panicking on error
- Fix R binding integer casts: add checked_u32 helper for i32→u32 inputs, change outputs from Vec<i32> to Vec<f64> to prevent truncation (10 new tests)


sanghoonio (approved):
good

PR #250, opened 2026-04-05
Align calc_gc_content, calc_dinucl_freq, and R type casts across bindings

sanghoonio:
## Summary

Combined finalization of binding alignment audit. Three related fixes:

### 1. Expose `calc_gc_content` + both `calc_dinucl_freq` variants in CLI

- Add `--fasta <PATH>` flag to `gtars genomicdist` subcommand
- Add `--ignore-unk-chroms` flag (default: `false`, matches Python)
- When `--fasta` is provided, output JSON gains three fields:
  - `gc_content: {mean, per_region}`
  - `dinucl_freq`: global dinucleotide counts
  - `dinucl_freq_per_region: {region_labels, dinucleotides, frequencies}`

### 2. Python: rename + add missing variant

- Renamed `calc_dinucleotide_frequency` → `calc_dinucl_freq` (matches core name)
- Added `calc_dinucl_freq_per_region` (was missing entirely)
- **Breaking change**: `calc_dinucleotide_frequency` no longer exists

### 3. R binding fixes

- **Naming bug fix**: `r_calc_dinucl_freq` was wrapping `calc_dinucl_freq_per_region` — R users calling by this name were silently getting per-region data instead of global counts. Renamed to `r_calc_dinucl_freq_per_region`. Added new `r_calc_dinucl_freq` wrapping the global variant. **Breaking change**.
- Standardized `ignore_unk_chroms` to optional (defaults `false`, matching Python)
- Type cast audit: changed `u32 → i32` to `u32 → f64` for coordinate/width/count values. Widths can exceed `i32::MAX` (2.1 Gbp) for concatenated/genome-scale regions. Return types change from integer to double vectors for `r_calc_widths`, `r_calc_nearest_neighbors`, `r_chromosome_statistics`.

## Deferred

WASM bindings for `calc_gc_content` / `calc_dinucl_freq` skipped: would require a `GenomeAssembly` wrapper + client-side FASTA loading (multi-GB), which is impractical for browser use cases.

## Test results

| Suite | Result |
|---|---|
| `cargo test -p gtars-genomicdist` | 261 passed |
| `pytest tests/` (gtars-python) | 146 passed |
| CLI smoke (hg38 FASTA) | ✓ GC mean 0.659, dinucl counts, per-region rows sum to 100% |

## Test plan

- [ ] Review R breaking change (renamed `r_calc_dinucl_freq`) — R users who previously called this were getting per-region output, now they get global counts. If R wrapper functions exist downstream, they need updating.
- [ ] Review Python breaking change (renamed `calc_dinucleotide_frequency`)
- [ ] Consider whether R type return changes (integer → double) need downstream R wrapper updates

### DIFF
--- a/gtars-cli/src/genomicdist/cli.rs
+++ b/gtars-cli/src/genomicdist/cli.rs
@@ -44,6 +44,18 @@ pub fn create_genomicdist_cli() -> Command {
                 .help("Path to open signal matrix TSV (enables cell-type open chromatin enrichment)"),
         )
         .arg(
+            Arg::new("fasta")
+                .long("fasta")
+                .required(false)
+                .help("Path to genome FASTA file (enables GC content + dinucleotide frequencies)"),
+        )
+        .arg(
+            Arg::new("ignore-unk-chroms")
+                .long("ignore-unk-chroms")
+                .action(clap::ArgAction::SetTrue)
+                .help("When computing GC content, skip regions on chromosomes not in the FASTA (default: error)"),
+        )
+        .arg(
             Arg::new("promoter-upstream")
                 .long("promoter-upstream")
                 .required(false)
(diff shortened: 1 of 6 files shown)


sanghoonio:
Superseded: folded into #249 as an additional commit. Keeping #249 as the single binding-alignment PR.

PR #254, opened 2026-04-11
Add spatial-arrangement stats and fix gaps() chrom_sizes signature

sanghoonio:
## Summary

- Adds four new per-file summary statistics to `GenomicIntervalSetStatistics` (`calc_inter_peak_spacing`, `calc_peak_clusters`, `calc_density_vector`, `calc_density_homogeneity`) with full Python / R / Wasm bindings and CLI exposure.
- Rewrites `gaps()` to require a `chrom_sizes` argument — fixes a pre-existing correctness bug where trailing gaps (last peak → chromosome end) could never be emitted. **Breaking API change**; 3 internal callers updated.
- Closes pre-existing binding asymmetries: Python gains `gaps()`, R and Wasm gain `cluster()`.

### Rust core (`gtars-genomicdist`)

| Function | Returns | Notes |
|---|---|---|
| `calc_inter_peak_spacing` | `SpacingStats` | mean / median / std / IQR / log_mean / log_std over `calc_neighbor_distances()`; NaN for empty / singleton / all-overlapping inputs |
| `calc_peak_clusters(radius_bp)` | `ClusterStats` | wraps existing `cluster()`; `n_clusters`, `n_clustered_peaks`, `mean_cluster_size` (averaged over size-`>`-1 clusters), `max_cluster_size`, `fraction_clustered` |
| `calc_density_vector(chrom_sizes, n_bins)` | `DensityVector` | dense **zero-padded** per-window counts, karyotypic order, with `chrom_offsets` for per-chromosome slicing — suitable for ML feature extraction |
| `calc_density_homogeneity(chrom_sizes, n_bins)` | `DensityHomogeneity` | `n_windows`, `n_nonzero_windows`, `mean_count`, variance, CV, classical Gini; delegates to `calc_density_vector` to avoid the zero-window correctness trap |

28 new Rust unit tests (10 for `gaps()`, 18 for the new stats).

### `gaps()` rewritten

`IntervalRanges::gaps` now takes `&HashMap<String, u32>` (required) and emits:

- **leading** gaps (0 → first peak)
- **inter-region** gaps
- **trailing** gaps (last peak → `chrom_size`) — ✨ new
- **full-chromosome** gaps for any chromosome in `chrom_sizes` with no regions, matching `bedtools complement` — ✨ new

Output is karyotypically ordered. Regions extending past `chrom_size` are clipped. The new signature subsumes the originally-planned `largest_gaps` feature, which is dropped — callers wanting top-N gaps can now filter `gaps()` output directly.

### Pre-existing binding gaps closed

| Function | Was in | Now also in |
|---|---|---|
| `gaps()` | R, Wasm | Python |
| `cluster()` | Python | R (as `clusterRegions`), Wasm |

### CLI

- `gtars ranges gaps` — new required `--chrom-sizes` flag.
- `gtars genomicdist` — four new top-level JSON fields: `inter_peak_spacing`, `peak_clusters` (array across stitching radii), `density_vector`, `density_homogeneity`. Density fields gated on `--chrom-sizes` (warning when skipped, matching `expected_partitions`).
- New `--cluster-radii` flag (default `"500,5000,50000"`) probes promoter / enhancer / domain scales in a single invocation — cheap multi-scale spatial fingerprint.

### Bindings

- **Python** (`gtars-python`): `PyRegionSet` gains `gaps()`, `inter_peak_spacing()`, `peak_clusters()`, `density_vector()`, `density_homogeneity()`. Four new `#[pyclass]` result wrappers with attribute getters and `__repr__`. `.pyi` stubs updated. 11 new pytest cases.
- **R** (`gtars-r`): new S4 generics `clusterRegions`, `interPeakSpacing`, `peakClusters`, `densityVector`, `densityHomogeneity`. `gaps()` method signature updated to take `chrom_sizes` (keeping `start`/`end` for `IRanges::gaps` generic compatibility, documented as ignored). `extendr-wrappers.R`, `NAMESPACE`, and 11 new `.Rd` man pages regenerated via `rextendr::document()`. 11 new testthat cases.
- **Wasm** (`gtars-js`): new methods on `JsRegionSet` (`cluster`, `interPeakSpacing`, `peakClusters`, `densityVector`, `densityHomogeneity`); `gaps()` signature updated. Results serialized via `serde_wasm_bindgen`. Compiles clean against `wasm32-unknown-unknown`.

### Side fix: `calcDinuclFreq` R wrapper arity

`rextendr::document()` surfaced a pre-existing bug: `R/extendr-wrappers.R` was checked in with a stale 3-arg signature for `r_calc_dinucl_freq`, while the Rust source has taken 4 args since commit `bd1e5b7`. The regenerated wrapper file is correct, and `R/genomicdist.R` is updated to pass the 4th `ignoreUnknownChroms` argument so `calcDinuclFreq()` dispatches without a runtime arity error. Fixes 2 previously-failing R tests.

## Scope notes

Dropped from the original plan:
- **`largest_gaps`** — subsumed by the `gaps()` fix (trailing gaps now included).
- **`chromosome_distribution`** — already derivable from `chromosome_statistics()`.

## Test plan

- [x] `cargo test -p gtars-genomicdist` — **296 passed** (+28 new)
- [x] `cargo check --workspace` — clean
- [x] `cargo check -p gtars-js --target wasm32-unknown-unknown` — clean
- [x] `pytest tests/test_regionset.py` — **25 passed** (+11 new)
- [x] `Rscript -e 'devtools::test()'` — **288 passed** (+11 new; 2 pre-existing `calcDinuclFreq` failures fixed by the side fix)
- [x] End-to-end CLI smoke against ENCODE ENCFF730DLS (23K regions, hg19): all four new JSON fields populate with scientifically plausible values (median inter-peak spacing ~17 kb, clustering fraction rising 53% → 71% → 89% across 500 bp / 5 kb / 50 kb scales, CV = 1.63, Gini = 0.69)
- [ ] CI (GitHub Actions) — will run on push

## Migration note for `gaps()` callers

Before:
```rust
let g = rs.gaps();
```
After:
```rust
let g = rs.gaps(&chrom_sizes);
```

Python / R / JS callers need to pass a `chrom_sizes` mapping. See updated docstrings and test cases for examples.


sanghoonio:
## Integration-test findings from region-cnn

Exercised all four new spatial-stats methods (`inter_peak_spacing`, `peak_clusters`, `density_vector`, `density_homogeneity`) plus the updated `gaps()` on 11 ENCODE narrowPeak files (~5K to 300K peaks each) using canonical hg38.chrom.sizes. End-to-end throughput was ~0.26 s/file and all methods returned sensible values — internal consistency checked, field values all reconstruct correctly. The core math is solid.

Three API/documentation issues worth addressing before merge. None are correctness bugs, all are ergonomic traps a caller can fall into. Listed in decreasing severity.

---

### 1. `n_bins` parameter name is misleading (medium)

`density_vector(chrom_sizes, n_bins)` and `density_homogeneity(chrom_sizes, n_bins)` both take `n_bins`, but that value is not the total number of bins you get back — it's the bin count for the longest chromosome, and every other chromosome gets proportionally fewer (rounded up). With canonical hg38.chrom.sizes (455 entries including alts):

```
n_bins=10  -> bin_width=24,895,642  total_windows=568
n_bins=100 -> bin_width=2,489,564   total_windows=1693
n_bins=249 -> bin_width=999,824     total_windows=3585
```

A user asking for \`n_bins=249\` reasonably expects 249 bins, not 3585. The actual semantic is "target bin count on the longest chromosome" or equivalently "inverse bin width, calibrated to the longest chromosome."

Suggestions (pick whichever fits the existing gtars naming conventions):
- Rename the parameter to something like \`bins_per_longest_chrom\`, \`ref_n_bins\`, or \`longest_chrom_bins\`.
- Or accept \`bin_width_bp\` as an alternative constructor parameter, since that's the quantity most users actually care about.
- At minimum, document the semantic prominently in both the Rust docstring and the `.pyi` stub. Current `.pyi` just says "n_bins" without elaborating.

### 2. `chrom_sizes` with alt contigs silently inflates `n_windows` (low-medium)

Related to #1 but has its own failure mode. Canonical UCSC hg38.chrom.sizes has 455 entries — 25 canonical chromosomes plus 430 alt/random/unplaced contigs. At `n_bins=249`, **408 of those 455 contigs get exactly 1 bin each** because they're smaller than `bin_width=999,824` bp. That inflates `n_windows` from ~3177 (canonical only) to 3585 and dilutes `density_homogeneity` stats: the mean count drops, Gini goes up, CV shifts, all from empty alt-contig bins that no caller genuinely cares about.

The bias is small for typical ChIP/ATAC peak sets (the empty bins add to the "zero counts" tail) but it's silent — there's no warning, and a naive caller loading `hg38.chrom.sizes` from UCSC and passing it in directly will get subtly different stats than they'd get if they pre-filtered to canonical chromosomes.

Suggestions:
- At minimum: document in both the Rust docstring and the `.pyi` stub that `chrom_sizes` should probably be filtered to chromosomes you care about, with an example of filtering out alt/random/chrUn.
- Optionally: add a small helper `canonical_chrom_sizes(chrom_sizes)` or similar that drops `_random`, `_alt`, `chrUn_*`, `chrEBV`, etc. Would be useful across gtars, not just for these new methods.
- Consider whether bins smaller than some minimum should be dropped entirely. Contigs shorter than `bin_width` arguably shouldn't appear in the density vector at all — right now they contribute a single bin that's usually empty.

### 3. `ClusterStats` field semantics mix "all clusters" with "multi-peak clusters only" (medium)

`n_clusters` counts all connected components including singletons, but `mean_cluster_size` is averaged **only over clusters of size > 1**. A naive caller computing `n_clusters * mean_cluster_size` as "total peaks in clusters" gets the wrong answer. Verified on encode_4 (105,678 peaks) at r=5000:

```
n_clusters         = 22,052   (includes singletons)
n_clustered_peaks  = 96,842   (peaks in size-> 1 clusters)
mean_cluster_size  = 7.328    (mean over size-> 1 clusters only)

naive product: 22052 * 7.328 = 161,596    (wrong — not 96,842, not 105,678)
correct:       n_multi = n_clusters - (total - n_clustered) = 22052 - 8836 = 13,216
               n_multi * mean_cluster_size = 13216 * 7.328 = 96,842 ✓
```

The info is recoverable but requires knowing the implementation detail (that singletons are counted in `n_clusters`). Suggestions:

- Add an `n_multi_clusters` field alongside `n_clusters`. It's free to compute and removes all ambiguity.
- **Or** rename `mean_cluster_size` to `mean_multi_cluster_size` so the asymmetry with `n_clusters` is surfaced in the field names.
- **Or** make `mean_cluster_size` average over all clusters (including singletons, which contribute size 1). This gives a different number but is less surprising relative to `n_clusters`. Downside: for files with many singletons, the mean gets pulled toward 1 and becomes less informative.

I'd lean toward adding `n_multi_clusters` as the cheapest fix that preserves the existing semantics for anyone already consuming it.

---

### What worked well

For the record, a few things that were especially nice to integrate against:

- `inter_peak_spacing`'s log_mean / log_std on log10(gap+1) — heavy-tailed gap distributions are a pain otherwise, and having the log-space stats precomputed saved me from rolling a numpy helper.
- `DensityVector.chrom_offsets` — correct karyotypic ordering and per-chromosome slice recovery without needing to re-bin. This is exactly the right shape.
- `gaps()` now includes trailing gaps and full-chromosome gaps — the `bedtools complement` parity is what I needed for any "largest gaps" followup.
- No errors across 11 files × 4 methods = 44 calls. The empty / singleton / all-overlapping edge cases the plan called out are handled correctly; I didn't find a single NaN or crash.

Will update this comment if more issues turn up when I run against the larger Rivanna corpus.

sanghoonio:
Thanks for the detailed review and the integration-test data — really useful to have concrete numbers from real ENCODE BEDs. Pushed two follow-up commits addressing all three points:

## Issue 1 (`n_bins` misleading) — [f533827](https://github.com/databio/gtars/commit/f533827)

**Docs-only fix.** Updated the docstrings on `calc_density_vector` and `calc_density_homogeneity` (and their binding equivalents across Python / R / Wasm / CLI help text) to clearly state the semantic: `n_bins` is the target bin count for the **longest chromosome** in `chrom_sizes`, not the total window count. The formula (`bin_width = max_chrom_len / n_bins`) is now spelled out, and there's a concrete recipe for computing `n_bins` from a desired bin width: `n_bins = max_chrom_len / desired_bin_width_bp`.

Didn't rename the parameter. Keeping it as `n_bins` preserves consistency with `region_distribution_with_chrom_sizes` (which uses the same name for the same quantity), and a rename would be a breaking change at the binding layer. If the doc clarity isn't enough in practice, a follow-up PR can add `calc_density_vector_by_width(chrom_sizes, bin_width_bp)` as a sibling entry point.

## Issue 2 (alt contigs silently inflate `n_windows`) — [f533827](https://github.com/databio/gtars/commit/f533827)

Also docs-only, same commit. Added a "Per-chromosome bin width" section to both trait docstrings and to the `DensityVector` / `DensityHomogeneity` struct-level docs. It explains:

- The last bin on each chromosome is narrower than `bin_width` whenever `chrom_size` isn't an exact multiple (this is true for all chromosomes, not just alts — the "short contig" case is just the extreme).
- Chromosomes shorter than `bin_width` (common with UCSC alt / random / unplaced contigs like `chrUn_*`, `*_random`, `*_alt`) reduce to a single bin whose effective width equals the chromosome size.
- `counts[i]` is a count per bin, not per-bp — bins of different effective widths are not directly comparable as densities when `chrom_sizes` includes contigs significantly shorter than `bin_width`.

Deliberately framed this as a technical description of what the function does, not as a "you should pre-filter your chrom_sizes" prescription. Users now have the information to decide whether to filter, set a bigger `n_bins`, or live with the diluted stats.

No runtime warning — warnings that fire on every canonical UCSC chrom.sizes file would just be noise.

## Issue 3 (`ClusterStats` field semantics) — [90514a0](https://github.com/databio/gtars/commit/90514a0)

This one turned into a real design discussion. I went a different route than your suggested `n_multi_clusters` field: instead of making the struct more complex, I made `min_cluster_size` apply **uniformly** to every size-dependent field in `ClusterStats` (except `max_cluster_size`, which is inherently a max-over-all-clusters concept).

**New semantics:**

| Field | With `min_cluster_size = 2` (default) | With `min_cluster_size = 1` |
|---|---|---|
| `n_clusters` | multi-peak clusters only | all connected components |
| `n_clustered_peaks` | peaks in multi-peak clusters | == `total_peaks` (tautology) |
| `mean_cluster_size` | mean of multi-peak cluster sizes | `total_peaks / n_clusters` (simple average) |
| `fraction_clustered` | peaks-in-multi / total | 1.0 (tautology) |
| `max_cluster_size` | biggest cluster overall | biggest cluster overall |

The identity `n_clusters * mean_cluster_size == n_clustered_peaks` now holds **at any threshold** — the naive-product trap is resolved not by adding a field but by eliminating the asymmetry that caused it. Verified on ENCFF730DLS (23,420 peaks):

```
default (min=2), r=5000:
  n_clusters=5614, n_clustered=16555, mean=2.949, fraction=0.707
  identity: 5614 * 2.949 == 16555 ✓

min=1, r=5000:
  n_clusters=12479, n_clustered=23420, mean=1.877, fraction=1.000
```

At default, every number is about multi-peak clusters. At `min=1`, every number is about all clusters. No more mixed populations.

**Default is 2**, matching the plan's original intent for this function (enhancer clustering, super-enhancer stitching — singletons usually aren't the object of study). `min_cluster_size=1` is still available when you want the simple-average view; I just moved it off the default because `n_clustered_peaks == total_peaks` and `fraction_clustered == 1.0` aren't useful outputs.

- **Python**: `rs.peak_clusters(radius_bp, min_cluster_size=2)`
- **R**: `peakClusters(rs, radius_bp, min_cluster_size = 2L)`
- **Wasm**: `rs.peakClusters(radius_bp, 2)` (no defaults in JS, users pass both)
- **CLI**: new `--cluster-min-size` flag, defaults to 2

Also added a `test_peak_clusters_mixed_min_3` case to cover `min_cluster_size > 2` for completeness.

---

## Verification

- `cargo test -p gtars-genomicdist` — **299 passed** (+3 new)
- `pytest tests/test_regionset.py` — **26 passed** (+1 new)
- `Rscript -e 'devtools::test()'` — **294 passed** (no regressions)
- `cargo check --workspace` — clean
- End-to-end CLI smoke against real ENCODE data — numbers check out, identity holds

The PR is ready for another look. Let me know if anything else turns up when you run against the Rivanna corpus.

sanghoonio:
you're welcome, @sanghoonio

sanghoonio:
## Followup integration test against 90514a0 — all three findings resolved

Re-synced region-cnn to the latest feature-branch tip, re-ran `compute_features.py` on the same 11 ENCODE test files (~0.27 s/file, zero errors), and re-executed the exploration notebook end-to-end. All three issues from the earlier comment are addressed.

### 1 & 2 — `n_bins` semantic + `chrom_sizes` alt-contig inflation (f533827)

Resolved via docstring clarification rather than API rename. The `DensityVector`, `DensityHomogeneity`, `RegionSet.density_vector`, and `RegionSet.density_homogeneity` docstrings now spell out: `n_bins` is the target bin count for the longest chromosome (not the total), `bin_width = max(chrom_sizes.values()) // n_bins`, total window count is `sum(ceil(size / bin_width))`, and short contigs each contribute a narrow single-bin entry that dilutes mean/variance/CV/Gini. The instruction to pre-filter `chrom_sizes` if that behavior is unwanted is explicit in the docstring.

Fine by me — the docs fix costs no caller-facing breakage and anyone grepping for `n_bins` in the `.pyi` now sees the full semantic inline. The original rename suggestion would have been more surface area for no compensating benefit.

### 3 — `ClusterStats` field-semantics asymmetry (90514a0)

Resolved with the new `min_cluster_size: int = 2` parameter on `peak_clusters`. This is cleaner than my suggested "add `n_multi_clusters` field" — one parameter covers both the default "size ≥ 2" view and the "include singletons" case, and the identity `n_clusters * mean_cluster_size == n_clustered_peaks` now holds at any threshold. Behavioral verification on encode_4 at r=5000:

```
min_cluster_size=2 (default):
  n_clusters         = 13,216
  n_clustered_peaks  = 96,842
  mean_cluster_size  = 7.328
  fraction_clustered = 0.916
  identity: 13216 * 7.328 = 96,842 = n_clustered_peaks  ✓

min_cluster_size=1:
  n_clusters         = 22,052
  n_clustered_peaks  = 105,678 (== total_peaks, by definition)
  mean_cluster_size  = 4.792
  fraction_clustered = 1.0 (== 1.0, by definition)
```

The old behavior (`n_clusters=22052`) is still recoverable via `min_cluster_size=1`, so downstream code that wanted the raw connected-component count isn't cut off.

### Impact on region-cnn Stage 0 feature matrix

Diffed the full 11×136 feature parquet against the pre-update version:

- **3 columns changed**: `C_cluster_r500_n_clusters`, `C_cluster_r5000_n_clusters`, `C_cluster_r50000_n_clusters`. All dropped (by mean ~77%, ~51%, ~21% respectively at the three radii — the singletons).
- **133 columns byte-identical**: everything else, including all `_n_clustered`, `_mean_size`, `_max_size`, `_fraction`, and all spacing/density/homogeneity columns.
- **Identity `n_clusters * mean_size == n_clustered` holds 11/11 files at every radius.**

No code changes required in `compute_features.py` — it calls `rs.peak_clusters(radius)` with no kwargs and picks up the new default. Added an inline comment documenting the reliance on the default.

Notebook updated: section 4's "derive n_multi_clusters from n_clusters − n_singletons" construction is gone (no longer needed — `n_clusters` is already the multi-cluster count at the default), and a new cell demonstrates the `min_cluster_size=2` vs `=1` comparison so anyone reading the notebook sees the tradeoff explicitly.

Ready to merge from the region-cnn integration-test perspective. No blockers.

PR #258, opened 2026-05-05
gtars-uniwig: tolerate BAM headers with VN-less @PG records

nsheff:
## Summary

- noodles 0.83 strict-parses `@PG` records and rejects any without a `VN:` tag, but SAM spec v1 §1.3 marks `VN:` as recommended (only `ID:` is required). PEPATAC/bowtie2/samtools BAMs routinely lack `VN:`, so `gtars uniwig` panics at signal generation.
- New `bam_header::read_header_lenient` reads BAM magic + `l_text` + raw header text, inserts `VN:unknown` into any `@PG` line missing it, and hands the sanitized text to noodles' standard `sam::Header` parser. Other invalid records still error.
- All four `reader.read_header().unwrap()` sites in `gtars-uniwig/src/lib.rs` swap to the helper. The indexed reader still handles `query()` — only the header parse is bypassed.

## Test plan

- [x] 12 unit tests for the sanitizer (CRLF, mixed records, substring false positives) and synthetic-BAM round-trip
- [x] 1 integration test (`tests/lenient_header.rs`) running `uniwig_main` on a fixture BAM with a hand-crafted VN-less `@PG`
- [x] Full `cargo test -p gtars-uniwig` green (76 tests)
- [x] Manual reproducer from the diagnosis guide runs to "FINISHED" instead of panicking


nsheff:
This may not be necessary after all... this was a red herring due to using an outdated version of bowtie2 which didn't record VN; problem solved by using a more recent version of bowtie2. So I'm not going to merge this after all.

PR #262, opened 2026-05-30
Indexed region set (1/2): additive core — IntervalSetOps + IndexedRegionSet

nsheff:
**Part 1 of 2** — split from #261 along its additive/destructive boundary.

This PR is **purely additive**: it adds the new indexing architecture alongside
the existing `IntervalRanges` code. Nothing is deleted, no caller changes, so it
can't break existing behavior — review it on the new code alone. The teardown and
migration are in #261 (stacked on this).

### What's added

**`3bac9158` — `IntervalSetOps` + structural methods (`gtars-core`); MCO source ownership**
- `SortedRegionSet` newtype
- Inherent ops on `RegionSet`: `reduce`, `concat`, `union`, `trim`, `shift`,
  `flank`, `resize`, `narrow`, `promoters`, `pintersect`, `disjoin`, `cluster`, `closest`
- `IntervalSetOps` trait (`setdiff`, `intersect`, `jaccard`, `coverage`,
  `overlap_coefficient`), sweep-line impls
- `MultiChromOverlapper`: owns an optional source `RegionSet` (`from_region_set()`,
  `source()`) + query methods with `min_overlap`

**`cdd8b425` — `IndexedRegionSet` wrapper**
- Pairs a `RegionSet` with a pre-built overlapper for build-once-query-many
- `count_overlaps` / `any_overlaps` / `find_overlaps` / `intersect_all`; `Deref` to
  `RegionSet`; 12 unit + 28 doc tests

### Verification
`cargo test` green for `gtars-core`, `gtars-overlaprs`, `gtars-genomicdist`
(40 + 277 + 76 + doctests). `--all-features` whole-workspace build needs R for the
`extendr` bindings; those crates aren't touched here.


sanghoonio:
**Two notes for keeping the additive core a clean drop-in for the part-2 migration: `closest` correctness + a missing `gaps`.**

Since this PR is the "additive" half and part 2 migrates genomicdist/consumers onto these core ops, two things are worth fixing *here* so the core surface is actually a faithful target — everything else (e.g. removing the `subtract` alias, strand ops) can wait for the genomicdist-side refactor.

**1. `RegionSet::closest` should be the genomicdist implementation, not a fresh copy.**
The new core `closest` (`gtars-core/src/models/region_set.rs:976`) diverges from the existing `IntervalRanges::closest` (`gtars-genomicdist/src/interval_ranges.rs:778`) in two ways:
- **Correctness:** it enumerates indices over a *sorted clone* of `other` (`region_set.rs:980`), so the returned `other_idx` points into that internal sorted copy rather than the caller's `other`. When `other` isn't already start-sorted, the index maps to the wrong region. (genomicdist captures `(idx, r)` *before* sorting, so it returns the original index.)
- **Performance:** it sorts `other` but doesn't use the sort to search — it still linear-scans every region per query, i.e. O(R·M). genomicdist binary-searches + expands with a best-distance bound, ~O(R·(log M + k)).

Demo test (temporary, delete after review): `gtars-genomicdist/tests/closest_divergence_demo.rs`
```
cargo test -p gtars-genomicdist --test closest_divergence_demo -- --nocapture
```
Output — both find distance 10, but the indices map to different regions:
```
--- self region: chr1 100-110 ---
core  -> other_idx=0 dist=10  => other.regions[0] = chr1 500-510   (wrong)
gdist -> other_idx=1 dist=10  => other.regions[1] = chr1 120-130   (true nearest)
```
Suggest: drop the core copy and port genomicdist's `closest` verbatim (it's pure — sort + binary search, no index dependency).

**2. `gaps` is missing from the core surface.**
`IntervalRanges::gaps` (`interval_ranges.rs:602`) — the per-chromosome complement bounded by chrom sizes (≈ `bedtools complement`) — isn't in the core ops. Both Python and wasm call it, so part-2 can't fully move them onto core until it exists here. It's a pure sweep op so it belongs in core; one snag to decide when porting: it currently sorts output via `crate::utils::chrom_karyotype_key` (`interval_ranges.rs:688`), so either move that helper down too or have core sort plainly by chr name.

No other ops need to be added in this PR — `subtract` (a one-line alias of `setdiff`) and the rest are better handled on the genomicdist side post-merge.

gtars-core/src/models/region_set.rs
```diff
@@ -648,6 +648,556 @@ impl RegionSet {
...
+                cluster_end = cluster_end.max(r.end);
+            }
+            result[idx] = cluster_id;
+        }
+
+        result
+    }
+
+    /// Find the nearest region in `other` for each region in `self`.
+    pub fn closest(&self, other: &RegionSet) -> Vec<(usize, usize, i64)> {
```
sanghoonio, line 1109: **Replace with genomicdist's `IntervalRanges::closest`.** This copy has a correctness bug — it returns indices into an internal *sorted clone* of `other` (line below), so `other_idx` points at the wrong region when `other` isn't pre-sorted — and is O(R·M) vs genomicdist's O(R·log M). See the top-level review comment + `tests/closest_divergence_demo.rs` for a reproduction.

gtars-core/src/models/region_set.rs
```diff
@@ -648,6 +648,556 @@ impl RegionSet {
...
+            end: current.end,
+            rest: None,
+        });
+
+        RegionSet::from(merged)
+    }
+
+    /// Combine two region sets without merging overlapping intervals.
+    pub fn concat(&self, other: &RegionSet) -> RegionSet {
+        let mut regions = self.regions.clone();
```
nleroy917, line 711: Just a thought... would it make sense to take ownership of `self` and then avoid a `.clone()`? Or maybe a secondary API called `concat_into` (or something)...? that lets you consume the `RegionSet`?

I can see the API commonly being something like:
```rust
let rs1 = RegionSet::from(...)?;
let rs2 = RegionSet::from(...)?;

let concatented = rs1.concat(rs2)?;
```

And I'll never use `rs1` r `rs2` again... so it could be nice to avoid a clone and just consum the region set being concatenated. Just an idea...

gtars-core/src/models/region_set.rs
```diff
@@ -648,6 +648,556 @@ impl RegionSet {
...
+        regions.extend(other.regions.iter().cloned());
+        RegionSet::from(regions)
+    }
+
+    /// Merge two region sets into a minimal non-overlapping set.
+    ///
+    /// Equivalent to `self.concat(other).reduce()`.
+    pub fn union(&self, other: &RegionSet) -> RegionSet {
+        self.concat(other).reduce()
```
nleroy917, line 732: Same here

nleroy917: Actually, nvm. `concat` has the `clone()` so it'd carry over here

gtars-core/src/models/region_set.rs
```diff
@@ -648,6 +648,556 @@ impl RegionSet {
...
+                start: r.start.saturating_sub(upstream),
+                end: r.start.saturating_add(downstream),
+                rest: None,
+            })
+            .collect();
+        RegionSet::from(regions)
+    }
+
+    /// Pairwise intersection of two region sets by index position.
+    pub fn pintersect(&self, other: &RegionSet) -> RegionSet {
```
nleroy917, line 1008: Suggestion to name `pw_intersect` since "pintersect" sounds funny when saying it aloud. But maybe thats a good thing :)

gtars-overlaprs/src/multi_chrom_overlapper.rs
```diff
@@ -425,6 +443,234 @@ pub fn build_indexed_overlapper(
     MultiChromOverlapper {
         index_maps: core,
         overlapper_type,
+        source: None,
```
nleroy917, line 446: Just curious why this was added? Is it useful for someting else?

gtars-core/src/models/region_set.rs
```diff
@@ -648,6 +648,556 @@ impl RegionSet {
...
+}
+
+// ── IntervalSetOps trait ────────────────────────────────────────────────
+
+/// Two-set interval operations on genomic region sets.
+///
+/// Provides set algebra (setdiff, intersect) and similarity metrics
+/// (jaccard, coverage, overlap_coefficient). Implementations may use
+/// sweep-line or index-based algorithms.
+pub trait IntervalSetOps {
```
nleroy917, line 1279: I see this is in `models/region_set.rs`. Small thing and it probably doesnt matter but just curious if this would be better suited for something like `genomicdist`?

nleroy917 (approved):

gtars-core/src/models/region_set.rs
nsheff: pintersect is the function name used in biocondcutor so it's what we were familiar with...

gtars-overlaprs/src/multi_chrom_overlapper.rs
nsheff: the MCO is now a view on top of a regionset, this points to the source regionset, right?

nsheff: in part 2 this actually gets removed; MCO becomes source-free, IndexedRegionSet becomes the only thing that owns a source. this is just temporary.

gtars-core/src/models/region_set.rs
nsheff: It's a single 6-method trait implemented by both RegionSet and the MCO, and after part 2 the MCO, which goes to gtars-overlaprs, implements it too. Both implementors sit at or below core in the dependency graph; genomicdist is downstream of them. Putting the trait in genomicdist would invert that  core and
  overlaprs would have to depend on genomicdist to implement it.

gtars-core/src/models/region_set.rs
nsheff: ok yes, updated.

